PR #31, opened 2015-01-01
Search: loop by track

peterlvilim:
I don't expect this to be merged.  I haven't tested manually or run the test suite.  So there is a good chance that it breaks functionality.  I figured this was the easiest way to illustrate the performance regression that I am observing.

For me this diff when trying to add the search results for Avicii to now playing reduces the add time from over a minute to under 1 second.

Let me know in my bug report if you can reproduce/agree on this problem, and I will look at getting together a pull request that could be merged.

### DIFF
--- a/mopidy_spotify/library.py
+++ b/mopidy_spotify/library.py
@@ -63,20 +63,13 @@ class SpotifyLibraryProvider(backend.LibraryProvider):
     def _lookup_artist(self, sp_link):
         sp_artist = sp_link.as_artist()
         sp_artist_browser = sp_artist.browse(
-            type=spotify.ArtistBrowserType.NO_TRACKS)
+            type=spotify.ArtistBrowserType.FULL)
         sp_artist_browser.load()
-        for sp_album in sp_artist_browser.albums:
-            sp_album_browser = sp_album.browse()
-            sp_album_browser.load()
-            if sp_album.type is spotify.AlbumType.COMPILATION:
-                continue
-            if sp_album.artist.link.uri in VARIOUS_ARTISTS_URIS:
-                continue
-            for sp_track in sp_album_browser.tracks:
-                track = translator.to_track(
-                    sp_track, bitrate=self._backend._bitrate)
-                if track is not None:
-                    yield track
+        for sp_track in sp_artist_browser.tracks:
+            track = translator.to_track(
+                sp_track, bitrate=self._backend._bitrate)
+            if track is not None:
+                yield track
 
     def _lookup_playlist(self, sp_link):
         sp_playlist = sp_link.as_playlist()


mopidy_spotify/library.py
```diff
@@ -63,20 +63,13 @@ def _lookup_album(self, sp_link):
     def _lookup_artist(self, sp_link):
         sp_artist = sp_link.as_artist()
         sp_artist_browser = sp_artist.browse(
-            type=spotify.ArtistBrowserType.NO_TRACKS)
+            type=spotify.ArtistBrowserType.FULL)
```
kingosticks: If you only do this change do you still get the performance speed-up without sacrificing the loss of fucntionality?

peterlvilim: Good idea.  Unfortunately it doesn't help.  The performance loss is do to the have to load individual albums: sp_album_browser.load() which has to be done if you want the original 2.x functionality.

peterlvilim:
Did some more investigation.  I think Avicii makes for a good worse case example of this since he has songs appearing across many albums.  Currently the two if-continue statements are located after the load is called on the album browser.  One simple optimization is to move these up to the top of of the for loop.  With this approach Avicii takes about 30 seconds or so to load (with no cache) which is imo still too slow, but others may differ.

A couple ideas:
1. Could return to the 1.x style of looping all the tracks for the artist.  This is very fast but doesn't return all the tracks for the artist that may appear on various collections.  I know there is currently a filter for "VARIOUS_ARTIST" however this doesn't filter for instance for the Tron Legacy soundtrack where the artist is listed as Daft Punk but there is a remix by Avicii.  The 1.x style matches well against how I listen to music but may not be good for others.  Or include a configuration option for 1.x style looping.

2. Could keep the 2.x style with the continue optimizations I mentioned above.  I think this is to slow especially since ncmpcpp just hangs without any feedback to the user as to why it is taking so long to load the artist (e.g. a counter showing how how many albums left to load).

3. Could parallelize this.  Currently utils in pyspotify is responsible for loading albums.  It blocks until albums are done loading.  Which means that all albums for an artist are loaded sequentially with blocking IO.  Could create a non-blocking version of load and then a blocking collection construct so that the albums could be loaded as a group (in paralllel) and the blocking only occurs at the group level (i.e. wait until all albums are loaded but don't load them serially).

4. Could asynchronously update the now playing list so that music starts playing right away even though albums are still being loaded.  This is messy if the user decides to clear the now playing list or edit it while tracks are being added plus I don't even know if this is viable.

5. Could add a configuration option to limit the number of albums looped through for an artist since I believe spotify returns the artists primary works and then collections after that.

I personally am most interested in trying the parallel version.  But I thought I'd post these ideas to get some feedback.

mopidy_spotify/library.py
jodal: I believe the results may be truncated if you use `FULL` here, while with `NO_TRACKS` you remove a lot of data from the original request, making sure you get all albums. I might be wrong, though.

jodal:
I added debug timing in commit 2314305be8ee82c9856cea2a121ff7ef1287f805, and did `rm -r ~/.cache/mopidy/spotify` before each test run where I searched for "avicii" in ncmpcpp and added the artist entry to the tracklist, resulting in 1316 tracks in the tracklist.

1. Original naive implementation: 82.7s
2. Commit 0b698a20f14687f29681c2b694d2d4035081c278: Only load album, not browser, before doing compilation/VA checks: down to 30.7s
3. Commit b314b77cb57f4afe12ebaf4a2761e4b59ecf61b2: Get all album browsers first, so libspotify is told what we're going to use soon, then load and use them all: 2.8-3.5s

From a performance perspective, I think this is approaching an acceptable load time, since Avicii is really a worst case scenario.

From a functional perspective, I think there are a lot of uninteresting music in the tracklist, given that I only added the artist Avicii. Either reverting to pyspotify 1.x behavior or finding better ways to filter out non-Avicii results is needed.

jodal:
With a recent fix to pyspotify, you can now run the following not use the libspotify cache and show debug timings from Mopidy-Spotify:

```
mopidy -o loglevels/mopidy_spotify=debug -o spotify/cache_dir=
```

kingosticks:
Does filtering on `sp_album.is_available` help at all? 
Is there any benefit in just checking `is_loaded` and getting on with doing the faster (smaller?) loading albums first and coming back to the slow (big) albums later?

mopidy_spotify/library.py
kingosticks: The libspotify docs say
> There is currently no built-in functionality available for getting the albums belonging to an artist. For now, just iterate over all tracks and note the album to build a list of all albums

Isn't using `FULL` the only way to get all the tracks? I know the `tophit` tracks are supposed to be missing but if there really are other results missing then that's hopeless. Probably worth looking into and verifying.  Although arguably you're potentially loading a lot of useless track data (various artists, compilations) that will just be filtered out. This kind of lookup is far superior in the web API; it's a shame they hate libspotify so much.

kingosticks:
And so I'm clear, the speed-up in the original post was presumably mostly due to libspotify's cache, rather than any change, right? Or did you manually clear the cache, @peterlvilim?

mopidy_spotify/library.py
trygveaa: That can't be right. sp_artist.browse (sp_artistbrowse_create) returns albums.

I thought it returned all albums when using FULL, but when I tried with Avicii now, it returned 100 albums, which seems to be a very round number, possibly indicating it has been cut off.

I tested with https://gist.github.com/trygveaa/6fc0244290d94223d540 (replacing input prompts with values).
NO_TRACKS: Took 1.5 seconds, returned 100 albums.
FULL: Took 7 seconds, returned 16 albums.

Also worth noting: The libspotify documentation says: "SP_ARTISTBROWSE_FULL: All information except tophit tracks. This mode is deprecated and will removed in a future release".

kingosticks: > That can't be right. sp_artist.browse (sp_artistbrowse_create) returns albums

That returns an `ArtistBrowser`.

I omitted the park about a future release because we all know that's never going to happen.

trygveaa: An `ArtistBrowser` which contains the albums.

kingosticks: Maybe they mean to say there's no single magic sp_artist function to just get albums (rather than all the other artist stuff too)? Or the note is just outdated.

jodal:
`.load()` when probably not needed just works as a guarantee that either:

- It is loaded. This is almost entirely free, since it just checks `.is_loaded` and returns.
- We load it now. The time we spend on loading it will also help all other existing `sp_*` objects to get loaded in the background.

kingosticks:
Inspired by @trygveaa's gist I had a little play around with spotipy (Spotify web api). I know there's no intention to use it alongside pyspotify (although it's search filtering is rather good so it may be worth revisiting this) - this was purely out of curiosity. Incidentally I love how easy pyspotify v2.x is to use! 

It's slightly relevant here since it demonstrates the different results you can get.  Also, the timings I got for both pyspotify and spotipy were both all over the place, varying from 0.2s to 5s on subsequent runs. No idea why.
https://gist.github.com/kingosticks/6c31f111fe116a76a85b

kingosticks:
Having played around with this again today I'd like to bump the idea of adding `sp_album.is_available` as another filter condition. It prevents a load of duplicate bogus albums which are just autolinked to the real available album. 

An example of this is the following album results where only 5xwKv8NCfinmOfLkEoRqSk is available (in my region) and the tracks in the unavailable albums autolink to it. I only noticed since the album name was slightly messed up for some of the unavailable albums.
```
Lost Forever / Lost Together (2014) - spotify:album:5xqZxgnnJ3ZY927zuBQhsN (spotify:artist:3ZztVuWxHzNpl0THurTFCv)
Lost Forever // Lost Together (2014) - spotify:album:5xwKv8NCfinmOfLkEoRqSk (spotify:artist:3ZztVuWxHzNpl0THurTFCv)
Lost Forever // Lost Together (2014) - spotify:album:0i4q4A0tytzQmYjgITsJAz (spotify:artist:3ZztVuWxHzNpl0THurTFCv)
Lost Forever, Lost Together (2014) - spotify:album:6SMsp6iS7YwaKSKtyIgiHI (spotify:artist:3ZztVuWxHzNpl0THurTFCv)
```

jodal:
I'm revisiting this now, so just adding some notes while I work.

1. Run: `mopidy -o loglevels/mopidy_spotify=all -o spotify/cache_dir=`
2. Search for 'avicii' in ncmpcpp
3. Add the "Avicii - Artist: Avicii" entry to the queue
4. Look at the artist lookup timing in the debug output

- Initially: 2270ms, 1101 tracks in queue
- Check album.is_available: 749ms, 298 tracks in queue

This bug was originally about the speed, which we already got down from ~83s to ~2s. With this change, we're further down to 0.75s, and the amount of results are a lot more manageable. Further, we had too much duplicates, which I hope this change makes a lot better.

Since this bug already encompasses two issues, both which are somewhat addressed, I'm closing it. Please retest and open new issues if we still have too much duplicates/uninteresting results in the artist lookup.

PR #37, opened 2015-01-20
Add Offline playlist support for spotify

richardambridge:
Hi again. Second attempt at pull request to introduce offline playlists for mopidy-spotify

PR #39, opened 2015-01-20
Set private session once logged in

richardambridge:
Only set the private session information once we have logged in to spotify


adamcik:
Same as https://github.com/mopidy/mopidy-spotify/pull/37#issuecomment-70657935

PR #40, opened 2015-01-20
Add Starred playlist

richardambridge:
Add the starred items playlist.

### DIFF
--- a/mopidy_spotify/playlists.py
+++ b/mopidy_spotify/playlists.py
@@ -38,11 +38,14 @@ class SpotifyPlaylistsProvider(backend.PlaylistsProvider):
         pass  # TODO
 
     def lookup(self, uri):
-        try:
-            sp_playlist = self._backend._session.get_playlist(uri)
-        except spotify.Error as exc:
-            logger.debug('Failed to lookup Spotify URI %s: %s', uri, exc)
-            return
+        if(uri.endswith("starred") ):
+            sp_playlist=self._backend._session.get_starred()
+        else:    
+            try:
+                sp_playlist = self._backend._session.get_playlist(uri)
+            except spotify.Error as exc:
+                logger.debug('Failed to lookup Spotify URI %s: %s', uri, exc)
+                return
 
         if not sp_playlist.is_loaded:
             logger.debug(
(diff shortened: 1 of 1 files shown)


mopidy_spotify/playlists.py
```diff
@@ -38,11 +38,14 @@ def delete(self, uri):
         pass  # TODO
 
     def lookup(self, uri):
-        try:
-            sp_playlist = self._backend._session.get_playlist(uri)
-        except spotify.Error as exc:
-            logger.debug('Failed to lookup Spotify URI %s: %s', uri, exc)
-            return
+        if(uri.endswith("starred") ):
+            sp_playlist=self._backend._session.get_starred()
```
kingosticks: What if I try to access someone else's starred playlist?

adamcik:
Same as https://github.com/mopidy/mopidy-spotify/pull/37#issuecomment-70657935

kingosticks:
In fact, it just needs to grab the username from the URI and use it in the call to `get_starred()`. Didn't see you could do that.

@jodal Is it worth allowing people to create a Playlist from a starred URI in pyspotify? At the moment that doesn't work as the `Link.as_playlist()` call fails (wrong LinkType). Else, any client always has to catch the special starred case.

kingosticks:
> @jodal Is it worth allowing people to create a Playlist from a starred URI in pyspotify? At the moment that doesn't work as the Link.as_playlist() call fails (wrong LinkType). Else, any client always has to catch the special starred case.

Fixed in https://github.com/mopidy/pyspotify/commit/94c362418582d717e0654970e815ca2d1a04f187

jodal:
As @kingosticks said, this has been fixed in pyspotify and will work without any changes in Mopidy-Spotify when the next pyspotify version is released.

kingosticks:
Did you mean to close this @jodal ? Don't you still need something in mopidy-spotify to call `get_starred()`?

jodal:
On Mon, Jan 26, 2015 at 4:22 PM, Nick Steel <notifications@github.com>
wrote:

> Did you mean to close this @jodal <https://github.com/jodal> ? Don't you
> still need something in mopidy-spotify to call get_starred()
>
Did you delete this comment, @kingosticks? *confused*

kingosticks:
Yep, guilty. Realised shortly after posting that it's all fixed and my question was unessesary. 

Sorry for the confusion, deleting comments is a bad idea, no matter how inane.

Issue #23, opened 2014-09-08
Make Spotify cover art available to clients


jodal:
Spotify cover art should be made available to clients.

Possible approaches ranged from simple/heavy to smarter/lighter:

- Add `data:` URIs with the covers to the `Album` objects. Not a good idea for clients already overloaded by Mopidy sending them too much data.
- Add `http://` URLs pointing to cover art hosted by a web extension provided by Mopidy-Spotify. This would be a lot better.
- Wait for Mopidy core to implement a generic metadata lookup service or a more specific cover art lookup service and provide Spotify cover art through this service. The related Mopidy issue is mopidy/mopidy#263.

jodal:
Mopidy 1.0 adds `library.get_images(uri)`, ref mopidy/mopidy#973.

#46 has been opened to add support for `get_images()` and replaces this issue.

PR #44, opened 2015-03-15
Add (temporary?) cover image support

vespakoen:
Hi!

I have made a `hacky?` patch that adds cover art support (returning the data_uri)
This can be displayed with `background: url('the_base64_from_data_uri')` and ``

I have read the discussion in https://github.com/mopidy/mopidy-spotify/issues/23
And I know a better solution is on it's way, but perhaps this can help other people out that want it right now.

### DIFF
--- a/mopidy_spotify/library.py
+++ b/mopidy_spotify/library.py
@@ -84,6 +84,7 @@ class SpotifyLibraryProvider(backend.LibraryProvider):
         sp_artist_browser.load()
         top_tracks = list(translator.to_track_refs(
             sp_artist_browser.tophit_tracks))
+
         albums = list(translator.to_album_refs(sp_artist_browser.albums))
         return top_tracks + albums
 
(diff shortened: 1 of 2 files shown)


adamcik:
Thanks for taking a stab at this. Sadly due to design mistakes we've made in Mopidy's APIs I don't think we can merge this, but I can offer an alternative. But firstly, lets quickly go over the mistakes blocking this:

1) Since we don't dedupe the model info you'll be adding the image data once per track for every track in an album with the same image.
2) We return fat models in a ton a places we shouldn't, search being the biggest offender.

Additionally there are also the `browse` and `lookup` code paths which I didn't see in this PR.

As for the alternative, luckily for you 0.20 which should get released any day now has https://docs.mopidy.com/en/develop/api/backends/#mopidy.backend.LibraryProvider.get_images which works around a lot of these problems. So any change concerning album art should be using this API.

For this this you have two approaches, do something like https://gist.github.com/adamcik/786643627eeae5620629 which uses pyspotify (but I'm unclear on the libspotify license with respect to cover art use), over even better cleanup https://gist.github.com/adamcik/a0b72ec356264d06aaf6 which uses the spotify web APIs.

So if you were to clean up the web based one, maybe add some tests and then plug that into `get_images` I think we have a winner which I would support merging :-)

adamcik:
Fixed via 74ac793d29ae34cfefe23de653a76a6393f36e51 which adds `get_images` support.

vespakoen:
Hooray!

Best news from today! Thanks a million!

PR #45, opened 2015-03-15
Adding helping out section to readme

vespakoen:
Might be helpful for future contributions?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -141,6 +141,28 @@ Project resources
 - `Download development snapshot <https://github.com/mopidy/mopidy-spotify/tarball/develop#egg=Mopidy-Spotify-dev>`_
 
 
+Helping out
+===========
+
+```
+git clone https://github.com/mopidy/mopidy-spotify.git
+cd mopidy-spotify
+
+# ### start editing files
+
+# install your changes
+sudo python setup.py install
+# restart mopidy service
+sudo service mopidy restart
+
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -141,6 +141,28 @@ Project resources
...
+
+The following commands should get you going.::
+
+    git clone https://github.com/mopidy/mopidy-spotify.git
+    cd mopidy-spotify
+
+    # start editing files
+
+    # install your changes
+    sudo python setup.py install
```
adamcik: You normally want `python setup.py develop` and not install, as develop will create a pointer to your dev dir so you don't have to re-install for every change. I would also recommend doing things in a virtualenv and not globally, but that might be out of scope for such a getting started guide.

vespakoen: Thanks for the insights, will update.
Not very familiar with python over here...

vespakoen:
Most commits ever for a simple change like this, i'm sorry =)

jodal:
With Mopidy 1.0, there's a new development guide that also applies to most extensions at https://docs.mopidy.com/en/latest/devenv/

I think a link to this guide is a good candidate for adding to the cookiecutter-mopidy-ext project template instead of adding instructions to every extension.

vespakoen:
Thanks, looks great!

Perhaps link to it from the readme?, that's how I ended up here ;)

PR #69, opened 2015-09-15
Add proxy support

### LINKED ISSUE #68: Proxy support
Proxy support was forgotten when reimplementing Mopidy-Spotify for the 2.0 release.

Ref. mopidy/pyspotify#176.

jodal:
Fixes #68

### DIFF
--- a/mopidy_spotify/backend.py
+++ b/mopidy_spotify/backend.py
@@ -89,13 +89,30 @@ class SpotifyBackend(pykka.ThreadingActor, backend.Backend):
     def _get_spotify_config(self, config):
         ext = Extension()
         spotify_config = spotify.Config()
+
         spotify_config.load_application_key_file(
             os.path.join(os.path.dirname(__file__), 'spotify_appkey.key'))
+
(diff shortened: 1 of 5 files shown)


mopidy_spotify/backend.py
```diff
@@ -89,13 +89,30 @@ def _get_session(self, config):
...
+
         if config['spotify']['allow_cache']:
             spotify_config.cache_location = ext.get_cache_dir(config)
         else:
             spotify_config.cache_location = None
+
         spotify_config.settings_location = ext.get_data_dir(config)
+
+        proxy_uri = None
+        if config['proxy'].get('hostname'):
```
adamcik: Why not use `httpclient.format_proxy`?

adamcik: That is with `auth=False`

jodal: That's a good idea... the current code was just stolen from Mopidy-Spotify 1.x.

jodal:
Updated PR.

Issue #72, opened 2015-10-04
In ncmpcpp, unable to enter the Album column


GregorySchwartz:
Using mopidy with mopidy-spotify, local albums work find and one can enter from the left column to the middle column in the ncmpcpp media library. However, if the middle column highlighted album contains a spotify album with a date of 0, e.g. |Jerry Goldsmith|(0) A Patch of Blue|BLANK, then you are stuck in the left column and cannot enter the middle column at all. In the mopidy output, you are left with an endless loop of
`
INFO     Searching Spotify for: album:"A Patch Of Blue" year:0 artist:"Jerry Goldsmith"
INFO     Searching Spotify for: album:"A Patch Of Blue" year:0 artist:"Jerry Goldsmith"
INFO     Searching Spotify for: album:"A Patch Of Blue" year:0 artist:"Jerry Goldsmith"
`
etc. However, as you are stuck in the left column, you can still scroll away, but it's very inconvenient. Is there a way to remove those seemingly non-existent albums from the middle column? I can try to figure this out and work on it, but I would need a starting point of where to look.

robin92:
I'm affected by this issue as well.

marcaddeo:
Also having this issue, will have to look into the code at some point and see if I can put in a PR. Very inconvenient though :cry:

jodal:
I'm not experiencing this issue with ncmpcpp 0.6.4. Are you running ncmpcpp 0.5.x?

jodal:
Even though I didn't get stuck, I saw the albums with year 0, which didn't make much sense, so I made those albums have an unknown date instead.

GregorySchwartz:
I am no longer experiencing this issue, thanks.

PR #74, opened 2015-10-10
Do not include 0 date albums for spotify searching

GregorySchwartz:
This commit solves issue #72 by not including spotify albums with a year of 0. This commit will obviously not include any of year 0, but the assumption is spotify (in my experience) will not have 0 years.

### DIFF
--- a/mopidy_spotify/search.py
+++ b/mopidy_spotify/search.py
@@ -44,7 +44,9 @@ def search(config, session, query=None, uris=None, exact=False):
     sp_search.load()
 
     albums = [
-        translator.to_album(sp_album) for sp_album in sp_search.albums]
+        translator.to_album(sp_album)
+        for sp_album in sp_search.albums
+        if sp_album.year != 0]
     artists = [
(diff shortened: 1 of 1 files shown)


marcaddeo:
@GregorySchwartz so this basically just ignores the 0-year albums completely from the index? That doesn't seem like the correct fix.

For instance, *Dub FX*'s album *A Crossworlds* has a year of 0 according to mopidy, but a year of 2014 according to the Spotify UI... So there seems to be some sort of disconnect on the "import"?

Disclaimer: I haven't gone in and learned how this thing works yet :smile:

GregorySchwartz:
@marcaddeo that's correct, this is just a stop gap until the real problem is fixed. Obviously we would want all albums that are indexed to be included, regardless of date, but this commit is intended to be a temporary workaround until the real issue is fixed.

jodal:
60848b698e98fc41df9fa690316c040e91b2b2fc fixes #72 by ignoring 0-dates in the translator, instead of the user of the translated data, like this PR does.

Issue #22, opened 2014-09-08
Add support for Spotify playlist management


jodal:
Moved from mopidy/mopidy#42:

> It would be nice to be able to manage Spotify playlists through Mopidy. E.g. create/rename/remove playlists, and add/move/remove tracks in playlists.

peterlvilim:
Was thinking of adding this feature.
- Create new spotify playlists
- Add and remove tracks from spotify playlists
- Rename spotify playlists

Right now I assume any playlist that you create would have to only include spotify items in order for it to sync to spotify.  I'm guessing this feature might also require additions to pyspotify which I could do as needed.

Planning to read through some of the code at which point I'll have something more intelligent to say on this topic.

charles-l:
:+1:

lino:
#80 should do the trick.

Cysioland:
Currently getting 'Playlist with scheme "spotify" can't add track with scheme "spotify"' error in ncmpcpp, which does not make sense.

jodal:
Please don't hijack unrelated issues.

Cysioland:
But it is related, because Mopidy-Spotify currently can't manage playlists,
hence this message.

pbogut:
Any update about this one?

yannik-b:
Not now, spotify will close libspotify, which is used by this project, until 2017 or it is closed already (https://github.com/mopidy/mopidy-spotify/issues/110)
So there will (proprably) be currently no changes for libspotify.

kingosticks:
There are no changes required to libspotify (or pyspotify) to implement this feature as they already have [full support for modifying playlists](https://pyspotify.mopidy.com/en/latest/api/playlist/). The changes required to implement this feature are in mopidy-spotify/mopidy. 

If I remember correctly the main blocker is some sensible generic handling when tracks come from multiple backends, as per @peterlvilim's comment above.

So, no update here sadly.

Radvendii:
I think ideally you could add tracks from multiple sources to a playlist in mopidy, and it would sync only those tracks that are from spotify back to the spotify playlist, no?

ecly:
Is there anything going on with this?

kingosticks:
I don't know of anyone working on it.

We need to start using the Web API for most playlist stuff so that probably
includes this. Although since you can't edit the Spotify owned playlists
(the ones seemingly now incompatible with libspotify) that might not
technically be the case. But doing everything the same way does have value.

[...]

PR #80, opened 2015-12-02
Implemented playlist save & delete

lino:
It's pretty shitty but it adds rudimentary playlist saving. (As requested e.g. in #22 )
Needs unit tests.

### DIFF
--- a/mopidy_spotify/playlists.py
+++ b/mopidy_spotify/playlists.py
@@ -93,7 +93,8 @@ class SpotifyPlaylistsProvider(backend.PlaylistsProvider):
         try:
             sp_playlist = self._backend._session.get_playlist(playlist.uri)
         except spotify.Error as exc:
-            logger.debug('Failed to lookup Spotify Playlist URI %s: %s', playlist.uri, exc)
+            logger.debug('Failed to lookup Spotify Playlist URI %s: %s',
+                         playlist.uri, exc)
             return
         if not sp_playlist.is_loaded:
             logger.debug(
@@ -103,7 +104,8 @@ class SpotifyPlaylistsProvider(backend.PlaylistsProvider):
             logger.debug("Renaming playlist %s to %s",
                          sp_playlist.name, playlist.name)
             sp_playlist.name = playlist.name
-        if len(sp_playlist.tracks) < len(playlist.tracks): # item was added
+        if len(sp_playlist.tracks) < len(playlist.tracks):
+            # item was added
             added_uri = None
             added_idx = None
             for idx, track in enumerate(playlist.tracks):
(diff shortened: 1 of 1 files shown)


jodal:
If nothing else (haven't looked at the code yet, except it's relative short length), this is a nice kick in the but to have me fix this once and for all. Thanks :-)

lino:
Please note that this will only work as long as only granular changes are processed. It will still need some work if multiple changes are to be saved at the same time.

jelly:
Just tested this functionality. Adding a new track to a new playlist seems to work in ncmpcpp, you end up with two playlists for the newly created playlist "mopidy" however: (as seen in ncmpcpp)

```
mopidy
mopidy [2]
```

Adding another song to the playlist also works. Adding a song from the "playlists browser" in ncmpcpp shows an error, but does add the track to the playlist.
```
Playlist with scheme "spotify" can't add track with scheme "spotify"
```

Deleting a song from the playlist also shows an error, but actually succeeds.

Renaming also works, but ncmpcpp's playlist view isn't being updated. I am not sure if this is a bug of mopidy or ncmpcpp though. (sonata also does not seem to update playlists)
Renaming a non-existant (playlist renamed in another client) does not crash mopidy

bartekb81:
Getting rid of those errors is quite simple: just make sure to return the playlist from the input param at the end of the *save* method:

```
        logger.info('Saved %s tracks to playlist %s',
                    len(playlist.tracks), playlist.name)
	return playlist
```

This is what save method in mopidy.core.playlists requires.

Everything works like a charm after that change.

bartekb81:
And a comment on delete() method. It didn't work on my installation, here's an adjusted version with comments:

```
    def delete(self, uri):
        try:
           logger.info(str(uri))
           container = self._backend._session.playlist_container
           if not container.is_loaded:
               container.load()
           for idx, instance in enumerate(container):
               if isinstance(instance, spotify.playlist.Playlist):  # @bartek1981: it can also be a PlaylistContainerFolder so we need to check
                   if instance.link.uri == uri:                     # @bartek1981: link.uri instead of uri
                       del container[idx]
                       break
        except spotify.Error:
            logger.warning('Could not delete Spotify playlist "%s"', uri)
```

bartekb81:
Another comment on the save method. The check that decides if a track has been added:

```
            if len(sp_playlist.tracks) < len(playlist.tracks): # item was added
```

may break if there are some tracks in *sp_playlist* that are no longer available on Spotify. *playlist* doesn't include such tracks but sp_playlist does so this comparison will not always be correct.

Solution: iterate through sp_playlist and ignore unavailable tracks when computing the count. Use that count instead:

```
            available_sp_playlist_tracks = filter(lambda track: track.availability == 1, sp_playlist.tracks)

            if len(available_sp_playlist_tracks) < len(playlist.tracks): # item was added
            .... # replace all sp_playlist.tracks with available_sp_playlist_tracks

            elif len(available_sp_playlist_tracks) > len(playlist.tracks): # item was removed
            .... # replace all sp_playlist.tracks with available_sp_playlist_tracks
            ....
```

jodal:
To simplify the development setup I removed the `develop` branch in preference for just using a single `master` branch. That had the unintentional consequence of closing all PRs which targeted the `develop` branch.

With the recent porting of Mopidy to Python 3, the current `master` branch has changed quite a bit, so this PR would have needed a rebase now anyway. Feel free to open a new PR based on the current `master` branch if you're still interested in this feature.

Issue #81, opened 2015-12-07
Spotify playlists not found when adding to the tracklist


kingosticks:
Originally reported at https://discuss.mopidy.com/t/mpc-cli-spotify-playlist-not-found-unless-recently-browsed-in-http-client/943

If you try to add a random spotify playlist, i.e. one that you do not follow, directly to the tracklist then only those tracks that are also present in your followed playlists will show up. If none of the playlist's tracks exist in your followed playlists then it'll fail (as in the example below). But if then you then try to add the same playlist again it will work. 

> nick@broom:~$ mpc add spotify:user:spotify:playlist:5WjDxcjetee4nnHt2T6fQl
error: directory or file not found
nick@broom:~$ mpc playlist | wc -l
0
nick@broom:~$ mpc add spotify:user:spotify:playlist:5WjDxcjetee4nnHt2T6fQl
nick@broom:~$ mpc playlist | wc -l
22

Steps to reproduce:
1. Clear / move away your spotify cache
2. Add another user's Spotify playlist to the tracklist which contains some tracks that are not present in your followed playlists
3. Observe the missing tracks in the tracklist
4. Repeat step 2 and now see all the tracks are present

It works the second time as the tracks are now loaded.

PR #82, opened 2015-12-07
Ensure a playlist's tracks are all loaded when doing a lookup

kingosticks:
This should fix the issue described in #81. 

Perhaps we should be calling load on tracks for all the lookup routines? 

I'm not sure what to do with the test. I tried to swap `sp_playlist_mock` for `sp_unloaded_playlist_mock` but that's not working at all.

### DIFF
--- a/mopidy_spotify/lookup.py
+++ b/mopidy_spotify/lookup.py
@@ -94,6 +94,7 @@ def _lookup_playlist(config, sp_link):
     sp_playlist = sp_link.as_playlist()
     sp_playlist.load()
     for sp_track in sp_playlist.tracks:
+        sp_track.load()
         track = translator.to_track(
             sp_track, bitrate=config['bitrate'])
         if track is not None:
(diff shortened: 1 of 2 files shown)


kingosticks:
I added something vaguely sane to the test. 

I now see that `sp_unloaded_playlist_mock` was always going to fail since the unloaded track it contains will not load. Maybe It'd be nice for `sp_unloaded_track_mock` to turn into `sp_track_mock` when load is called but maybe that's OTT.

jodal:
> Perhaps we should be calling load on tracks for all the lookup routines?

Related considerations like this is always what makes me defer clicking merge on obviously good bug fixes, because what if there's a similar bug right next to it that needs fixing to, but isn't that obvious to users so it doesn't get reported?

Looking through the similar code:

- `_lookup_track()` loads the track
- `_lookup_album()` loads the album browser, implicitly loading all tracks
- `_lookup_artist()`  loads the artist browser, then each album's browser, implictly loading all tracks
- `_lookup_playlist()` loads the playlist only, which will not load the tracks

So I think this fixes the only bug of this type. The rest looks good.

PR #96, opened 2016-02-14
Use Spotifys Web API for distinct

trygveaa:
Previously, distinct searched directly with pyspotify when given a query argument. This changes it to use the internal search method, which has been changed to using Spotifys Web API.

### DIFF
--- a/mopidy_spotify/distinct.py
+++ b/mopidy_spotify/distinct.py
@@ -4,40 +4,41 @@ import logging
 
 import spotify
 
-from mopidy_spotify import translator
+from mopidy_spotify import search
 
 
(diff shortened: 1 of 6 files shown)


jodal:
Thanks :-)

trygveaa:
Forgot to mention this, but since the album model for search in the Web API doesn't include artists or date, _get_distinct_albumartists and _get_distinct_dates will only return empty results.

A way to solve this could be to use https://developer.spotify.com/web-api/get-several-albums/ to look up the extra information.

jodal:
Can you update the changelog with a note about it?

trygveaa:
Done.

PR #109, opened 2016-07-17
Add config parameter to specify the market the search is run against

wiibaker:
This should bypass the situation when the user is not on US (default) market and the intended results are not available on US market. Should also then fix the issue if the results are only available on US market, but not on users market making the track unplayable.


kingosticks:
This was fixed in https://github.com/mopidy/mopidy-spotify/pull/131/commits/8eb583efed2e818292474c771de9b70714b59e26

PR #119, opened 2016-11-05
search: Don't quote the any field of search queries

trygveaa:
If you include quotes around the query, you only get exact matches. From
the API documentation:

    Keywords will be matched in any order unless surrounded by double
    quotation marks: q=roadhouse&20blues will match both "Blues
    Roadhouse" and "Roadhouse of the Blues" while q="roadhouse&20blues"
    will match "My Roadhouse Blues" but not "Roadhouse of the Blues".

This means that non-exact matches previously wasn't possible. I think we
should default to non-exact matches. With this, exact matches are still
possible by including "" around the query yourself.

With the non-exact match you can also use different query types at the
same time in the any field, e.g. artist and track name. This is not
possible with the exact match.

The other fields still use quotes for now. If you don't include quotes,
only the first word will be for that field, while the rest will be
treated like the words in the any field. It seems like if you want
multiple words for a specific field, you need to include the field name
in front of each word, e.g.: track:roadhouse%20track:blues

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+v3.x.x (UNRELEASED)
+-------------------
+
+- Don't quote the any field of search queries. This allows you do get non-exact
+  matches. Exact matches can still be done by including quotes ("") in the
+  search query yourself. (PR: #119)
+
 v3.0.0 (2016-02-15)
 -------------------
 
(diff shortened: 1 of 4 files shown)


kingosticks:
Supporting exact matches was historically always something we wanted [to do](https://github.com/trygveaa/mopidy-spotify/blob/ce50f1f9dbc3c60afc2c847e86e75610dbb00ea4/mopidy_spotify/search.py#L24) so you've got to appreciate the irony. Does it make any sense to pass `search.search()`'s `exact` flag into the translator and make the quotes dependent on that?

trygveaa:
Heh, I forgot about that. Yeah, that definitely makes sense, I'll change it.

PR #121, opened 2016-11-08
Include adding the apt.mopidy.com's GPG key and APT repo

j10sanders:
If users don't follow the instructions on https://docs.mopidy.com/en/latest/installation/debian/ and just run: sudo-apt-get install mopidy-spotify, then they will receive the following error: "Unable to locate package mopidy-spotify".

If you would prefer, I could change the line with the link to https://docs.mopidy.com/en/latest/installation/debian/ to include something like "Be sure to visit this URL to ensure you install the GPG key and APT repo before installing mopidy-spotify.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -51,6 +51,8 @@ Installation
 Debian/Ubuntu/Raspbian: Install the ``mopidy-spotify`` package from
 `apt.mopidy.com <http://apt.mopidy.com/>`_::
 
+    wget -q -O - https://apt.mopidy.com/mopidy.gpg | sudo apt-key add -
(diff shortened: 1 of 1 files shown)


kingosticks:
What you've added there still won't work without running `sudo apt-get update` too. Better to just link to the relevant docs since it'll save maintaining it in two places.

jodal:
Also, it'll requires us to maintain a copy of the difference between the jessie and upcoming stretch dist, just like we've until quite recently had to cover the difference between wheezy and jessie in the APT repo docs.

Issue #130, opened 2017-02-01
Add support for authenticated calls to the Spotify Web API


kingosticks:
Spotify [recently announced](https://developer.spotify.com/news-stories/2017/01/27/removing-unauthenticated-calls-to-the-web-api/) the following:
> On 29th May 2017 the /search /tracks /albums /artists and /users (and related) endpoints will start requiring an access token, bringing them in-line with all other Web API endpoints

We currently access the `/search` and `/tracks` endpoints so we'll soon need to use Spotify's [Authorization Code Flow](https://developer.spotify.com/web-api/authorization-guide/#authorization_code_flow) to continue doing so. The good news is that the infrastructure for doing this has already been setup and is used by [mopidy-spotify-web](https://github.com/lfcabend/mopidy-spotify-web). Merging mopidy-spotify-web and mopidy-spotify together (https://github.com/mopidy/mopidy-spotify/issues/114) makes even more sense and might be the first place to start.

We would then be able to start using the Web API for more of the features currently provided by libspotify and reduce our dependency on that deprecated library (https://github.com/mopidy/mopidy-spotify/issues/110). Except, of course, for streaming which is unfortunately still only possible through libspotify. 

The Web API also give us access to features/improvements that are not available in libspotify, such as:
* 'Your Music' (https://github.com/mopidy/mopidy-spotify/issues/16).
* 'Top Lists' / Discover Weekly / Release Radar (https://github.com/mopidy/mopidy-spotify/issues/122)
* Playlist images (https://github.com/mopidy/mopidy-spotify/issues/100)
* Better searching (https://github.com/mopidy/mopidy-spotify/issues/11)

tomkel:
The features that stem from the proposed merge would be greatly appreciated.

adamcik:
I started on https://github.com/mopidy/mopidy-spotify/tree/feature/oauth based on some other OAuth things I've done before. But I've not had time to fix the existing tests or add new ones. But this should be enough to survive the end-points that should soon require auth.

So if someone else has time to finish this up we could get a new release out in time for the API change. And then longer term someone could consider #114 with respect to this.

theos-space:
I think the API change happened... my search refuses to work (just happened today!)

relthyg:
Can confirm the API change. `https://api.spotify.com/v1/search?type=artist&q=Opeth` returns:

```
{
  "error" : {
    "status" : 401,
    "message" : "No token provided"
  }
}
```

[...]

Issue #106, opened 2016-06-22
Can't find certain artist using mopidy-spotify 


Jorick:
I can't find the artist "The black box revelation" wen I use mopidy-spotify. 
Searching them with the Android app, or the Spotify webplayer works fine. So they are on Spotify.

I tried searching both with ncmpcpp en gmpc as frontends.
Searching for this band worked fine before the switch to the web api. I just haven't gotten around to report it.

I ran mopidy with the verbose flag enabled but it showed no errors.

PJkeeh:
Same here, also some albums are gone.

kingosticks:
In the case of @Jorick's "The black box revelation" this seems to be related to the Web API's `market` query parameter. We currently don't specify this parameter and apparently the default is 'US' where I guess the artist is not available on Spotify:
* [No market](https://api.spotify.com/v1/search?q=%22The+black+box+revelation%22&limit=50&type=album%2Cartist%2Ctrack) - 0 results
* [US market](https://api.spotify.com/v1/search?q=%22The+black+box+revelation%22&limit=50&type=album%2Cartist%2Ctrack&market=US) - 0 results
* [GB market](https://api.spotify.com/v1/search?q=%22The+black+box+revelation%22&limit=50&type=album%2Cartist%2Ctrack&market=GB) - many results!

This is fixed in https://github.com/mopidy/mopidy-spotify/pull/109 which just needs a bit of tidying up.

PJkeeh:
Can't wait

Jorick:
Sweet, thanks!

kingosticks:
Closing as this is really just a duplicate of https://github.com/mopidy/mopidy-spotify/issues/97. #109 was never merged but it looks like we won't want to fix it that way anyway.

PR #131, opened 2017-02-03
Fix missing / unavailable tracks in search results

### LINKED ISSUE #97: Search is returning unavailable tracks
With the migration to the spotify web api, the search results contains tracks that are not available in the user's country.

I found the @trygveaa comment on #94 about this, but we need a way to specify the desired market as a parameter or at least have the 'available_markets' array on the track/album returned object so we can filter it later. 

An exemple, as I'm living in Brazil (BR), https://api.spotify.com/v1/search?q=%22Menina+da+saia+verde%22&type=album%2Cartist%2Ctrack&limit=50

kingosticks:
This fixes #97 by setting the market parameter when doing searches using the Web API.

We don't currently set the parameter and it defaults to 'US' (https://github.com/spotify/web-api/issues/194). Depending on the logged in user's country, this may result in missing or unavailable tracks being returned. For example "The black box revelation" as reported in https://github.com/mopidy/mopidy-spotify/issues/106.

I appreciate that considering https://github.com/mopidy/mopidy-spotify/issues/130, this might not see the light of day, but at least this should remind us to set it as part of those future changes.

### DIFF
--- a/mopidy_spotify/search.py
+++ b/mopidy_spotify/search.py
@@ -56,6 +56,7 @@ def search(config, session, web_client,
     result = web_client.get(_API_BASE_URI, params={
         'q': sp_query,
         'limit': search_count,
+        'market': session.user_country,
         'type': ','.join(types)})
 
     albums = [
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -366,6 +366,7 @@ def session_mock():
     sp_session_mock = mock.Mock(spec=spotify.Session)
     sp_session_mock.connection.state = spotify.ConnectionState.LOGGED_IN
     sp_session_mock.playlist_container = []
+    sp_session_mock.user_country = 'GB'
     return sp_session_mock
 
 
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -77,6 +77,7 @@ def test_search_returns_albums_and_artists_and_tracks(
         params={
             'q': '"ABBA"',
             'limit': 50,
+            'market': 'GB',
             'type': 'album,artist,track'})
 
     assert 'Searching Spotify for: "ABBA"' in caplog.text()
@@ -124,6 +125,7 @@ def test_sets_api_limit_to_album_count_when_max(
         params={
             'q': '"ABBA"',
             'limit': 6,
+            'market': 'GB',
             'type': 'album,artist,track'})
 
     assert len(result.albums) == 6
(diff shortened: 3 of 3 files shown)


mopidy_spotify/search.py
```diff
@@ -55,11 +55,18 @@ def search(config, session, requests_session,
...
         search_count = 50
 
+    params = {
+        'q': sp_query,
+        'limit': search_count,
+        'type': ','.join(types)}
+
+    try:
+        params['market'] = session.user_country
+    except ValueError as exc:
```
trygveaa: Can this ever happen?

kingosticks: I don't really know, but remember, this is libspotify....

kingosticks: I guess if we were really worried about libspotify returning non utf-8 bytes back to us we should handle it in pyspotify. I can remove this (and it's test).

trygveaa: Hm, in libspotify it is an int and pyspotify doesn't raise any errors no matter what the value is, so I don't see how anything could happen. If the int is 0, you will get null-bytes, but that shouldn't trigger a ValueError.

kingosticks: If that returned int is greater than 65535, then `utils.to_country` will raise ValueError when it tries to convert a number greater 255 to `chr`. That routine could do with another `& 0xff` in there.

kingosticks:
Squashed and I've also removed the exception handling from this PR.  It'd only be needed if libspotify started misbehaving (which we have zero reports of), in which case it should be handled in pyspotify rather than here.

jodal:
Is this necessary now when all requests are authenticated? Couldn't Spotify know from your credentials which market you're in?

kingosticks:
You'd hope so, but not quite. From https://developer.spotify.com/web-api/search-item/:

> If from_token is given and a valid access token is supplied in the request header, only items with content playable in the country associated with the user's account will be returned. (The country associated with the user's account can be viewed by the user in their account settings at https://www.spotify.com/se/account/overview/). Note that the user must have granted access to the user-read-private scope when the access token was issued.

If you don't specify anything, you used to default to US but I'm not sure if that's still true.

adamcik:
> If market=from_token is supplied in the query string, the current user must have authorized the user-read-private scope.

So yeah, we do need the scope even if we never directly look it up.

jodal:
And @kingosticks needs to update this PR to apply to develop again :-)

kingosticks:
Well that didn't work

kingosticks:
That's better!

PR #135, opened 2017-03-21
Update README in case Spotify website breaks on device password

carlosdagos:
I had an issue when creating a device-specific password, because the Spotify website kept crashing. I suspect it's because I have a facebook-only login, and I suspect I'm not alone on this one. I managed to get everything working by following the steps outlined in the updated text.


jodal:
Thanks!

Issue #129, opened 2017-01-25
Media library timeout is always 10.000


eyadsibai:
if I modify the timeout in the config this is not being reflected. still I get timeout after 10 secs when I try to load the media library for example.

kingosticks:
We only seem to use the configured timeout for loading a track before we try and play it. Every other use of `load` is called with no arguments and hence uses pyspotify's default timeout of 10 seconds. 

Lots of people do hit this but it's usually a sign that sometime else is wrong (i.e. something at Spotify's end) since a 10 second timeout is really quite long. But we should fix it, good spot.

jaedb:
A good example of a genuine use case where 10s is not enough is when we want to play a playlist URI. 10 seconds would suffice for getting single objects, but a playlist lookup may be 100+ tracks.

For example: https://github.com/jaedb/Iris/issues/36

eyadsibai:
spotify-mopidy

oblitum:
Can't browse and play any playlist from Genres and Mods, always timing out.

oblitum:
Well, even after applying #139, no luck trying to browse Genres & Moods playlists...

kingosticks:
@oblitum you are seeing https://github.com/mopidy/mopidy-spotify/issues/133

jodal:
Fixed by PR #139.

PR #139, opened 2017-05-09
Make load calls respect timeout

mattfredwill:
As per #129, `load` calls were not respecting user configured timeout.


kingosticks:
Sorry for the delay, thank you for this.

PR #143, opened 2017-06-01
OAuth support for accessing Web API endpoints

kingosticks:
Fixes and tests for @adamcik's original branch.

Still need to consider (from #130):

- [x] check for id/secret from spotify_web, though that would mean making the new ones optional?

### DIFF
--- a/mopidy_spotify/web.py
+++ b/mopidy_spotify/web.py
@@ -1,0 +1,81 @@
+from __future__ import unicode_literals
+
+import logging
+import time
+
+import requests
+
+from mopidy_spotify import utils
+
+logger = logging.getLogger(__name__)
+
+
+class Error(Exception):
+    pass
+
+
+class OAuthClient(object):
+
+    def __init__(self, refresh_url, client_id=None, client_secret=None,
+                 proxy_config=None, expiry_margin=60):
+
+        if client_id and client_secret:
+            self._auth = (client_id, client_secret)
+        else:
+            self._auth = None
+
+        self._refresh_url = refresh_url
+
+        self._margin = expiry_margin
+        self._expires = 0
+
+        self._headers = {'Content-Type': 'application/json'}
+        self._session = utils.get_requests_session(proxy_config or {})
+
+    def _request(self, method, url, **kwargs):
+        try:
+            resp = self._session.request(method=method, url=url, **kwargs)
+            resp.raise_for_status()
+            return resp
+        except requests.RequestException as e:
+            raise Error('Fetching %s failed: %s' % (url, e))
+
+    def _decode(self, resp):
+        try:
+            return resp.json()
+        except ValueError as e:
+            raise Error('JSON decoding %s failed: %s' % (resp.request.url, e))
+
+    def _should_refresh_token(self):
+        return not self._auth or time.time() > self._expires - self._margin
+
+    def _refresh_token(self):
+        logger.debug('Fetching OAuth token from %s', self._refresh_url)
+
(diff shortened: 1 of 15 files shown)


mopidy_spotify/web.py
```diff
@@ -75,7 +75,14 @@ def _refresh_token(self):
...
-        kwargs.setdefault('headers', {}).update(self._headers)
-        return self._request('GET', url, **kwargs)
+        result = {}
+        try:
+            if self._should_refresh_token():
+                self._refresh_token()
+            kwargs.setdefault('headers', {}).update(self._headers)
+            response = self._request('GET', url, **kwargs)
+            result = self._decode(response)
+        except Error as e:
```
adamcik: We should probably catch a more specific error? Also do we want to log this just as debug?

kingosticks: I was thinking could raise and catch `OAuthTokenRefreshError` and a `OAuthClientError` exceptions. And I agree, these want more visibility, error level doesn't seem unreasonable to me.

kingosticks:
I'm done here, any comments on the latest commits? Happy to remove/edit the warning message if anyone has a better idea.

adamcik:
Not sure if caring about the credentials between the versions is the same makes any sense? Other wise this looks fine :-)

kingosticks:
I don't follow, between versions of mopidy-spotify? Which credentials? 

I was thinking maybe it should check the credentials provide the particular scopes we need? Did mopidy-spotify-web previously request whatever scope is needed for search and images or did you have to add that?

adamcik:
As in `spotify.client_* != spotify_web.client_*` probably not really mattering. If that clarifies things a bit.

For the scopes we currently request enough, but as soon as we start moving away from libspotify more whole sale this will start to matter more. And since there are already a bunch of tokens in use we just have to figure out how to migrate people somewhat nicely when the time comes.

kingosticks:
I thought it was necessary for both extensions to use the same set of credentials.  Shouldn't each Spotify user only be using one set of client_id/secret credentials at a time? If not, then I'll just revert that last commit.

adamcik:
You can have as many tokens a you like, and it makes no difference if there are separate ones for the two plugins. What could work is having `spotify` use the one from `spotify_web` if it's own value isn't set. But I think it makes more sense to make the `spotify` one required forcing people to fix this.

adamcik:
Other thing we could do is look over the scopes and reconfigure the defaults the oauthclientbridge requests, potentially reducing the migration required later?

kingosticks:
> You can have as many tokens a you like, and it makes no difference if there are separate ones for the two plugins. 

OK. I'm going to revert that last commit and squash the rest.

> What could work is having spotify use the one from spotify_web if it's own value isn't set. But I think it makes more sense to make the spotify one required forcing people to fix this.

Agreed.

> Other thing we could do is look over the scopes and reconfigure the defaults the oauthclientbridge requests, potentially reducing the migration required later?

I think this would be worthwhile. Getting playlists through the Web API is probably the next most pressing thing to fix so making sure we have all the `playlist-*` scopes would be nice.

kingosticks:
Ready to go! Do we need @jodal to make and upload a release?

adamcik:
I should have access at least, but haven't done this in ages. I'm kinda liking the idea of automating away all of this for everything in the mopidy org, but that is well beyond the scope of this PR. 

I'll go ahead and merge this as step one at least, and I might have time/energy to look at doing a release this evening.

pgrenaud:
Good job on this PR. 👏 I'd love to see a release today! 😄

PR #147, opened 2017-06-21
API cleanups to web client and adding retries

adamcik:
If this looks sane I would like to get this in before merging it into my other lookup related branch. Decided against lumping it in with the other branch so we can focus on the retries, the backoff etc.


adamcik:
Looks like the test failure is GST related :/

adamcik:
Updated to better handle URL merging like I was talking about on IRC. Also increase the backoff factor a bit. And finally made us shortcircuit after getting 401. All of the rate-limiting I've seen on the auth bridge is from clients that get 401s and then keep trying again, which this should solve.

Issue #146, opened 2017-06-20
When starting up pauses every other Spotify client


simmel:
When mopidy with mopidy-spotify is started it pauses every other Spotify client (even other mopidy-spotifies) even though mopidy isn't autostarting the music.

kingosticks:
I cannot reproduce this between instances of mopidy-spotify or my Android client. Only actually playing music causes the play token to be lost.

simmel:
Hm, I can't reproduce this as easy as I thought. But it has happened
enough times for me to get so annoyed I reported it at least ; )

I'll update when I can reproduce it 100%.

trygveaa:
Does Mopidy start in paused or stopped state when this happens?

kingosticks:
Actually Yeh I forgot about that relatively new queue/state restoring functionality. Are you using that?

simmel:
Mopidy was in paused state.
I don't think I am using any queue restore? You mean that the previously used.

It just happened now:
1. My kids were listening with the Spotify application on an iOS device
1. I resumed my laptop from sleep
    1. mopidy was already started but was in paused state
    1. WiFi reconnects
    1. mopidy takes a few minutes to realize that it has been disconnected from Spotify. In the logs I see: `INFO     Logged in to Spotify in online mode`
1. The iOS device pauses the music
1. mopidy on my laptop is still quiet and still paused.
1. I go to the iOS device and see that it displays "Music has been paused because the Spotify account is used on another device" (or whatever the text says).
1. I play on the iOS device
1. When I get back to my laptop I see in the mopdiy logs: `WARNING  Spotify has been paused because your account is being used somewhere else.`.

I'm now running mopidy with `-vv` to have more info when it happens again.

kingosticks:
Ahh right so it's because it's paused (rather than stopped) and restoring the connection (rather than starting mopidy). I think this is because to get to the paused state we need to go through the play state. A more verbose log would hopefully confirm that.

simmel:
That didn't say much:
```
INFO     2017-06-24 10:21:02,611 [11623:SpotifyEventLoop] mopidy_spotify.backend
  Logged in to Spotify in online mode
```
Gonna run it with three `-v`.

simmel:
Hopefully this makes more sense to you than me because I don't see anything obvious:

```
[...]
DEBUG    2017-06-24 11:06:18,180 [13986:Dummy-9] spotify.session
  Audio buffer stats requested, but no listener
[...]

PR #162, opened 2017-10-10
Pause spotify playback when Mopidy paused

kingosticks:
Previously, when paused, we stopped consuming audio buffers rather than telling libspotify to pause playback. This lead to undesirable behaviour on other clients if the "paused" mopidy-spotify instance was offline at the time playback was started on another Spotify client, causing it to miss the playback token lost signal from the other client (see #146).

### DIFF
--- a/mopidy_spotify/playback.py
+++ b/mopidy_spotify/playback.py
@@ -98,6 +98,11 @@ class SpotifyPlaybackProvider(backend.PlaybackProvider):
         self.backend._session.player.pause()
         return super(SpotifyPlaybackProvider, self).stop()
 
+    def pause(self):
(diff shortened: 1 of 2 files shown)


kingosticks:
As I said in #146, I'm unsure what effect (if any) this will have on ensuring we have continued to buffer sufficient data to allow us to instantly resume playback. libspotify stops sending us any more audio data as soon as we pause which is in contrast to the old behaviour which kept consuming the data until the buffer was full. I have not noticed any issues like this. 

But I have noticed the cpu usage is now slightly less when paused (max ~2.4% vs max ~0.7%). So that's a nice bonus.

PR #175, opened 2018-03-29
fixing problem of non-existence of 'images' attribute on album

felipemarinho97:
now album image support trough dBus and mPris interfaces are working fine

### DIFF
--- a/mopidy_spotify/translator.py
+++ b/mopidy_spotify/translator.py
@@ -68,10 +68,17 @@ def to_album(sp_album):
     else:
         date = None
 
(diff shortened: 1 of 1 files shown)


jodal:
The `Album.images` field was removed from Mopidy's models in Mopidy 3, which Mopidy-Spotify now require, so this PR is no longer relevant.

Mopidy-MPRIS has supported the `core.library.get_images()` method of getting cover art since v2.0 release almost a year ago, so the goal of this PR should already have been achieved that way.

PR #188, opened 2018-06-11
OAuthClient now cache header aware.

kingosticks:
This branch is still a work in progress but any thoughts are very welcome.

PR #195, opened 2018-08-18
Web client caching

kingosticks:
This adds caching to the web client and is required for decent performance when using Spotify's Web API for playlist data (https://github.com/mopidy/mopidy-spotify/issues/182, https://github.com/mopidy/mopidy-spotify/pull/188).

This supports expiry times calculated from the [Cache-Control header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) as well as [ETags](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) to detect unmodified response data once it's expired.

This part of the fix is self-contained and it seemed reasonable to split it out. I think I've also improved the code in doing so. Hopefully this will also make reviewing easier.

### DIFF
--- a/mopidy_spotify/web.py
+++ b/mopidy_spotify/web.py
@@ -154,16 +154,11 @@ class OAuthClient(object):
                 logger.debug('Fetching %s failed: %s', prepared_request.url, e)
                 status_code = None
                 backoff_time = None
-                expires = None
                 result = None
             else:
                 status_code = response.status_code
                 backoff_time = self._parse_retry_after(response)
-                expires = self._parse_cache_control(response)
-                etag = self._parse_etag(response)
-                json = self._decode(response)
-                result = WebResponse(
-                    prepared_request.url, json, expires, etag, status_code)
+                result = WebResponse.from_requests(prepared_request, response)
 
             if status_code >= 400 and status_code < 600:
                 logger.debug('Fetching %s failed: %s',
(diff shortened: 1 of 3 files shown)


kingosticks:
Rebased following pytest fixes in https://github.com/mopidy/mopidy-spotify/commit/4c9ef7ab538824e0cd4a41148cbc02ccf2629066

kingosticks:
@jodal so in my head this was merged but seems not. And that's maybe for the best as lumping the caching in with the oauth functionality isn't very nice. I'd be happy to pull it out into a CachingClient subclass or a CachingMixin. Or we could just get it in as is.

mopidy_spotify/web.py
```diff
@@ -128,11 +154,16 @@ def _request_with_retries(self, method, url, *args, **kwargs):
...
                 result = None
             else:
                 status_code = response.status_code
                 backoff_time = self._parse_retry_after(response)
-                result = self._decode(response)
+                expires = self._parse_cache_control(response)
+                etag = self._parse_etag(response)
+                json = self._decode(response)
+                result = WebResponse(
+                    prepared_request.url, json, expires, etag, status_code)
```
jodal: The `_parse_*` methods doesn't use `self` at all, so they can be placed anywhere.

I suggest keeping `WebResponse.__init__()` as is, but add a classmethod `WebResponse.from_requests(request, response)` that does parsing, etc before calling `__init__()` to create the object.

That change alone moves quite a lot of HTTP code from `OAuthClient` to `WebResponse`.

kingosticks: Good idea, done.

kingosticks:
Is this good to go, do we think?

PR #217, opened 2019-03-26
Replace yaourt with yay on arch linux installation steps

Eduruiz:


trygveaa:
Oh, I just did this change. Didn't see that there already was an open PR for it. I'll close this, since the change has been done.

PR #224, opened 2019-06-28
Fix backend tests compatibility with pykka 2.0.

kingosticks:
When asserting `mock.call in list(some_call)`, the `some_call` ends up on the lhs.
In our failing test cases, the `some_call` on the lhs includes an `ActorProxy` and the `mock.call` on the rhs includes the `mock.ANY`, so we get `ActorProxy.__eq__(mock.ANY)`, [which returns False](https://github.com/jodal/pykka/blob/develop/pykka/_proxy.py#L186-L188).
In Pykka 1.x there was no `ActorProxy.__eq__()` defined so we ended up doing `mock.ANY.__eq__(ActorProxy)`, [which returned True](https://github.com/python/cpython/blob/2.7/Lib/test/_mock_backport.py#L1880). This PR avoids this situation.

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -140,13 +140,12 @@ def test_on_start_adds_connection_state_changed_handler_to_session(
 
     get_backend(config).on_start()
 
-    assert (mock.call(
+    session.on.assert_any_call(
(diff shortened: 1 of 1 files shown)


jodal:
:+1:

PR #227, opened 2019-09-07
Use latest travis/tox receipes

kingosticks:
Use the default Travis build environment, which is now Xenial rather than Trusty (Python 2.7.12).


kingosticks:
Restarted the build a couple of times, it passed.

PR #234, opened 2019-11-19
Ensure status code is valid before using it.

kingosticks:
Python 3 fix: '>=' not supported between instances of 'NoneType' and 'int'

### DIFF
--- a/mopidy_spotify/web.py
+++ b/mopidy_spotify/web.py
@@ -170,7 +170,7 @@ class OAuthClient:
                 backoff_time = self._parse_retry_after(response)
                 result = WebResponse.from_requests(prepared_request, response)
 
-            if status_code >= 400 and status_code < 600:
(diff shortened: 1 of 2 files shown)


kingosticks:
Ooops. Failed formatting and a slight issue with the backoff sleep time making things slow.

kingosticks:
There we go

mopidy_spotify/web.py
```diff
@@ -170,7 +170,7 @@ def _request_with_retries(self, method, url, *args, **kwargs):
                 backoff_time = self._parse_retry_after(response)
                 result = WebResponse.from_requests(prepared_request, response)
 
-            if status_code >= 400 and status_code < 600:
+            if status_code and status_code >= 400 and status_code < 600:
```
jodal: This could be rewritten as:

```python
if status_code and 400 <= status_code < 600:
```

PR #228, opened 2019-09-12
Web API playlists v2

kingosticks:
This branch is the successor to #188 and aimed at fixing #140 and #182.

- [x] Cache playlist API response data
- [x] Support Spotify's new playlist URI scheme
- [x] ~~Save cached data to file on exit and restore on login~~ - not in this PR
- [ ] Translator should use @memoized decorator?
- [x] Track translators need to consider market availability
- [x] Tests
- [x] Handle playlists refresh

I'm not happy with returning the raw (mutable) cache data dict to `SpotifyOAuthClient` and this led to the hacky workaround at https://github.com/kingosticks/mopidy-spotify/blob/fix/web-api-playlists-v2/mopidy_spotify/web.py#L440-L443. Returning something immutable like a `namedtuple` might be better.

Issue #140, opened 2017-05-09
Cannot play playlists created by "Spotify"


corneboele:
We use the load by URI a lot, but recently we cannot load the playlist that are created by Spotify itself. Every other playlist or artist URI works perfectly. Any idea what could be the problem?

kingosticks:
This looks like https://github.com/mopidy/mopidy-spotify/issues/133

NiPKIT:
I have the same issue. I can see all my personal playlists and playlist from other people who i am following. But i cant see official playlists created by spotify.

poikilotherm:
Hi there,

I struggle from the same issue. Playlist built by myself or others are no problem, but anything related to "spotify:user:spotify:playlist:xxxx" is just not loading.

I didn't manage to enable debugging output for the extension - is this builtin?

Willing to help, debug and create a PR if someone can give me a hand.

jakuboz:
Hi,
same issue here for quite a long time. Just burned a fresh copy of raspbian and issue persists.
Playlists that are not created by user 'Spotify' work normally.
If I can provide any more details- I'll be happy to help.

kingosticks:
This issue is due to a breaking change in Spotify's libspotify library which is end of life and so will not be fixed by Spotify.  Therefore, the issue will persist until someone changes mopidy-spotify to use the Spotify Web API for playlist data instead of libspotify. 

I did some [initial work on this](https://github.com/kingosticks/mopidy-spotify/tree/fix/incompatible_playlists) and found that caching will be required, and of course a fair amount of tests. I won't have the time to work on it any further until after the holidays but others are more than welcome to have a go.

scheleaap:
I use [Iris](https://github.com/jaedb/Iris) in combination with Mopidy and Mopidy-Spotify to play Spotify playlists (at least my Discover Weekly playlist) and that works fine. Does that help you?

See https://github.com/scheleaap/ansible-role-mopidy/blob/master/tasks/install.yml if you're interested in the details of how I set up Mopidy.

NiPKIT:
I also use iris. In iris the spotify playlists work fine. I think iris uses [spotipy](https://github.com/plamere/spotipy) to interact with spotify. But when I use the command line (mpc) to list the playlists it doesn't work. I think because mopidy uses the old libspotify like @kingosticks mentioned.

jasondaigo:
[...]

PR #235, opened 2019-12-03
Web API playlists v3

kingosticks:
Third time lucky here. This is the Python 3 version of https://github.com/mopidy/mopidy-spotify/pull/228 aimed at fixing #140 and #182. Everything discussed in the previous PR is probably still valid for this version.

- [x] Cache playlist API response data
- [x] Support Spotify's new playlist URI scheme
- [x] ~~Save cached data to file on exit and restore on login~~ - not in this PR
- [ ] Translator should use @memoized decorator?
- [x] Track translators need to consider market availability
- [x] Tests
- [x] Handle playlists refresh
- [x] Include performance info

I have redone the performance measurements that I included in the commits for #228 which helped explained the code's evolution.

I'm *still* not happy with returning the raw (mutable) cache data dict to `SpotifyOAuthClient` and this led to the hacky workaround: https://github.com/mopidy/mopidy-spotify/blob/681debe77d46d233f792c7e64b1c76fc7bfff71e/mopidy_spotify/web.py#L462-L465. Returning something immutable like a `namedtuple` might be better or something Python 3-ish (dataclass?).

### DIFF
--- a/mopidy_spotify/playlists.py
+++ b/mopidy_spotify/playlists.py
@@ -6,7 +6,6 @@ import spotify
 
 from mopidy_spotify import translator, utils
 
-_cache = {}
 _sp_links = {}
 
 logger = logging.getLogger(__name__)
@@ -33,7 +32,7 @@ class SpotifyPlaylistsProvider(backend.PlaylistsProvider):
             return
 
         web_client = self._backend._web_client
-        for web_playlist in web_client.get_user_playlists(_cache):
+        for web_playlist in web_client.get_user_playlists():
             playlist_ref = translator.to_playlist_ref(
                 web_playlist, web_client.user_id
             )
@@ -62,14 +61,13 @@ class SpotifyPlaylistsProvider(backend.PlaylistsProvider):
 
     def refresh(self):
         with utils.time_logger("Refresh Playlists", logging.INFO):
-            _cache.clear()
             _sp_links.clear()
-            # Want libspotify to get track links so they load in the background
-            count = 0
-            for playlist_ref in self._get_flattened_playlist_refs():
-                self._get_playlist(playlist_ref.uri)
-                count = count + 1
-            logger.info(f"Refreshed {count} playlists")
+            with self._backend._web_client.refresh_playlists():
+                count = 0
+                for playlist_ref in self._get_flattened_playlist_refs():
+                    self._get_playlist(playlist_ref.uri)
+                    count = count + 1
+                logger.info(f"Refreshed {count} playlists")
 
         self._loaded = True
 
@@ -84,11 +82,11 @@ class SpotifyPlaylistsProvider(backend.PlaylistsProvider):
 
 
 def playlist_lookup(session, web_client, uri, bitrate, as_items=False):
-    if web_client is None:
+    if web_client is None or web_client.user_id is None:
         return
 
     logger.debug(f'Fetching Spotify playlist "{uri}"')
-    web_playlist = web_client.get_playlist(uri, _cache)
+    web_playlist = web_client.get_playlist(uri)
 
     if web_playlist == {}:
         logger.error(f"Failed to lookup Spotify playlist URI {uri}")
--- a/mopidy_spotify/playlists.py
+++ b/mopidy_spotify/playlists.py
@@ -3,7 +3,6 @@ import logging
 from mopidy import backend
 
 import spotify
-
 from mopidy_spotify import translator, utils
 
 _sp_links = {}
--- a/mopidy_spotify/playlists.py
+++ b/mopidy_spotify/playlists.py
@@ -113,13 +113,3 @@ def playlist_lookup(session, web_client, uri, bitrate, as_items=False):
                 logger.info(f"Failed to get link {track.uri!r}: {exc}")
 
     return playlist
-
-
-def on_playlists_loaded():
-    # Called from the pyspotify event loop, and not in an actor context.
-    logger.debug("Spotify playlists loaded")
-
-    # This event listener is also called after playlists are added, removed and
-    # moved, so since Mopidy currently only supports the "playlists_loaded"
-    # event this is the only place we need to trigger a Mopidy backend event.
-    backend.BackendListener.send("playlists_loaded")
--- a/mopidy_spotify/playlists.py
+++ b/mopidy_spotify/playlists.py
@@ -13,9 +13,13 @@ class SpotifyPlaylistsProvider(backend.PlaylistsProvider):
     def __init__(self, backend):
         self._backend = backend
         self._timeout = self._backend._config["spotify"]["timeout"]
+        self._loaded = False
 
     def as_list(self):
-        with utils.time_logger("playlists.as_list()"):
+        with utils.time_logger("playlists.as_list()", logging.INFO):
+            if not self._loaded:
+                return []
+
             return list(self._get_flattened_playlist_refs())
 
     def _get_flattened_playlist_refs(self):
@@ -34,11 +38,14 @@ class SpotifyPlaylistsProvider(backend.PlaylistsProvider):
                 yield playlist_ref
 
     def get_items(self, uri):
-        with utils.time_logger(f"playlist.get_items({uri})"):
+        with utils.time_logger(f"playlist.get_items({uri})", logging.INFO):
+            if not self._loaded:
+                return []
+
             return self._get_playlist(uri, as_items=True)
 
     def lookup(self, uri):
-        with utils.time_logger(f"playlists.lookup({uri})"):
+        with utils.time_logger(f"playlists.lookup({uri})", logging.DEBUG):
             return self._get_playlist(uri)
 
     def _get_playlist(self, uri, as_items=False):
@@ -47,7 +54,15 @@ class SpotifyPlaylistsProvider(backend.PlaylistsProvider):
         )
 
     def refresh(self):
-        pass  # TODO: Clear/invalidate all caches on refresh.
+        with utils.time_logger("Refresh Playlists", logging.INFO):
+            _cache.clear()
+            count = 0
+            for playlist_ref in self._get_flattened_playlist_refs():
+                self._get_playlist(playlist_ref.uri)
+                count = count + 1
+            logger.info(f"Refreshed {count} playlists")
+
+        self._loaded = True
 
     def create(self, name):
         pass  # TODO
@@ -63,7 +78,7 @@ def playlist_lookup(web_client, uri, bitrate, as_items=False):
     if web_client is None:
         return
 
-    logger.info(f'Fetching Spotify playlist "{uri}"')
+    logger.debug(f'Fetching Spotify playlist "{uri}"')
     web_playlist = web_client.get_playlist(uri, _cache)
 
     if web_playlist == {}:
--- a/mopidy_spotify/playlists.py
+++ b/mopidy_spotify/playlists.py
@@ -2,9 +2,12 @@ import logging
 
 from mopidy import backend
 
+import spotify
+
 from mopidy_spotify import translator, utils
 
 _cache = {}
+_sp_links = {}
 
 logger = logging.getLogger(__name__)
 
@@ -50,12 +53,18 @@ class SpotifyPlaylistsProvider(backend.PlaylistsProvider):
 
     def _get_playlist(self, uri, as_items=False):
         return playlist_lookup(
-            self._backend._web_client, uri, self._backend._bitrate, as_items
+            self._backend._session,
+            self._backend._web_client,
+            uri,
+            self._backend._bitrate,
+            as_items,
         )
 
     def refresh(self):
         with utils.time_logger("Refresh Playlists", logging.INFO):
             _cache.clear()
+            _sp_links.clear()
+            # Want libspotify to get track links so they load in the background
             count = 0
             for playlist_ref in self._get_flattened_playlist_refs():
                 self._get_playlist(playlist_ref.uri)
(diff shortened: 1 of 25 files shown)


kingosticks:
I redid the performance numbers and added the to the commit messages.

mopidy_spotify/playlists.py
```diff
@@ -1,90 +1,74 @@
...
+            self._backend._session,
+            self._backend._web_client,
+            uri,
+            self._backend._bitrate,
+            as_items,
         )
 
     def refresh(self):
-        pass  # Not needed as long as we don't cache anything.
+        if not self._backend._web_client.logged_in:
```
kingosticks: Should I also set `self._loaded = False` above this?

girst:
I've tried running your branch, but my playlists don't show up. I do have the `[Spotify]` entry with the top artists/albums/tracks inside, but none of my normal playlists, or the special 'my songs' and starred playlist.

<details><summary>installation method</summary>

1. install libspotify with prefix=/usr
2. cloned mopidy master; installed with setup.py
3. cloned this branch mopidy-spotify; installed w/ setup.py

`pip3 list|grep '[Mp]o[pt]'` then shows

```
Mopidy          3.0.0a6
Mopidy-Spotify  4.0.0a1
pyspotify       2.1.1
```

</details>

my .config/mopidy/mopidy.conf has these entries under `[spotify]`:
 - username
 - password
 - client_id
 - client_secret

what am i doing wrong?

kingosticks:
I've no idea what the "manual method" is. We only support one method as far as I know, and its the one outlined in this project readme. If you run Mopidy with verbose logging that would probably be useful.

Those attribute errors are a different issue, they have a bug on here already. They won't be the problem here.

I also suggest you check that `mopidy config` and `mopidy deps` agree with what you think is going on.

girst:
On Sat, Dec 07, 2019 at 09:40:59AM -0800, Nick Steel wrote:
>I've no idea what the "manual method" is. 

sorry, I've edited the previous comment: i forgot that this was the way 
to do it in the mopidy-spotify-web extension. you'd register a new 
application on developer.spotify.com and manually copy out the client 
details. I've been using them sucessfully with @blacklight's 
mopidy-spotify fork, so i assume they will work with this branch too?

>Those attribute errors are a different issue

i figured so.

>I also suggest you check that `mopidy config` and `mopidy deps` agree 
>with what you think is going on.

a log of mopidy --verbose and then navigating to the 'browse' view on 
ncmpcpp is here: https://paste.ubuntu.com/p/8stwW2w2Wv/ (i can't see 
anything)

mopidy config output: https://paste.ubuntu.com/p/KPPW22kq6b/

mopidy deps output: https://paste.ubuntu.com/p/KSpGY5zmw2/

girst:
ok, i've redone everything in a fresh VM (fc30), and now it works. sorry for the noise.

code itself looks good. i personally don't see anything wrong with the deepcopy 'hack' and would just leave it.

mopidy_spotify/web.py
```diff
@@ -356,24 +358,114 @@ def __str__(self):
 
 
 class SpotifyOAuthClient(OAuthClient):
+
+    TRACK_FIELDS = (
+        "next,items(track(type,uri,name,duration_ms,disc_number,track_number,"
+        "artists,album,is_playable,linked_from.uri))"
+    )
```
jodal: So... they've reinvented GraphQL inside REST? :man_facepalming:

mopidy_spotify/web.py
```diff
@@ -356,24 +358,114 @@ def __str__(self):
...
             refresh_url="https://auth.mopidy.com/spotify/token",
             client_id=client_id,
             client_secret=client_secret,
             proxy_config=proxy_config,
         )
+        self.user_id = None
+
+    def get_all(self, path, *args, **kwargs):
+        while path is not None:
+            logger.debug(f'Fetching page "{path}"')
```
jodal: Replace `"{path}"` with `{path!r}`.

mopidy_spotify/web.py
```diff
@@ -356,24 +358,114 @@ def __str__(self):
...
+            for page in pages:
+                for playlist in page.get("items", []):
+                    yield playlist
+
+    def get_playlist(self, uri, cache=None):
+        try:
+            parsed = parse_uri(uri)
+            if parsed.type != "playlist":
+                raise ValueError(
+                    f"Could not parse {repr(uri)} as a Spotify playlist URI"
```
jodal: Replace `{repr(uri)}` with `{uri!r}`.

mopidy_spotify/web.py
```diff
@@ -356,24 +358,114 @@ def __str__(self):
...
+        if more_tracks:
+            # Take a copy to avoid changing the cached response.
+            playlist = copy.deepcopy(playlist)
+            playlist.setdefault("tracks", {}).setdefault("items", [])
+            playlist["tracks"]["items"] += more_tracks
+
+        return playlist
+
+
+WebLink = collections.namedtuple("WebLink", ["uri", "type", "id", "owner"])
```
jodal: Replace with a proper dataclass?

``` python
from dataclasses import dataclass
from typing import Optional

@dataclass
class WebLink:
    uri: str
    type: str  # This could also be an enum
    id: Optional[str] = None
    owner: Optional[str] = None
```

With this, you could also move `parse_uri()` to a classmethod `WebLink.from_uri(uri)`.

mopidy_spotify/translator.py
```diff
@@ -149,14 +153,18 @@ def to_track_refs(sp_tracks, timeout=None):
...
-    # TODO: Check availability
+    # Web API track relinking guide says to use original URI.
+    # libspotfy will handle any relinking when track is loaded for playback.
+    uri = web_track.get("linked_from", {}).get("uri") or web_track["uri"]
 
-    return models.Ref.track(
-        uri=web_track.get("uri"), name=web_track.get("name")
-    )
+    if not web_track.get("is_playable", False):
+        logger.warning(f"{uri} is not playable")
```
jodal: Replace `{uri}` with `{uri!r}`?

mopidy_spotify/playlists.py
```diff
@@ -47,7 +54,15 @@ def _get_playlist(self, uri, as_items=False):
         )
 
     def refresh(self):
-        pass  # TODO: Clear/invalidate all caches on refresh.
+        with utils.time_logger("Refresh Playlists", logging.INFO):
```
jodal: Replace "Refresh playlists" with "playlists.refresh()" for consistency with the other time loggers?

mopidy_spotify/playlists.py
```diff
@@ -47,7 +54,15 @@ def _get_playlist(self, uri, as_items=False):
...
+            _cache.clear()
+            count = 0
+            for playlist_ref in self._get_flattened_playlist_refs():
+                self._get_playlist(playlist_ref.uri)
+                count = count + 1
+            logger.info(f"Refreshed {count} playlists")
```
jodal: INFO level messages should have enough context to make sense without the logger name, so replace "playlists" with "Spotify playlists".

mopidy_spotify/playlists.py
```diff
@@ -85,12 +94,31 @@ def playlist_lookup(web_client, uri, bitrate, as_items=False):
...
+        else:
+            tracks = playlist.tracks
+
+        for track in tracks:
+            if track.uri in _sp_links:
+                continue
+            try:
+                _sp_links[track.uri] = session.get_link(track.uri)
+            except ValueError as exc:
+                logger.info(f'Failed to get link "{track.uri}": {exc}')
```
jodal: Replace `"{track.uri}"` with `{track.uri!r}`?

mopidy_spotify/playlists.py
jodal: If we were able to process new requests while we refresh, it would make sense to continue serving the playlists we already have for concurrent requests while refreshing. Setting `self._loaded = False` would make those imaginary requests get an empty response instead. Or have I missed something?

kingosticks: I was thinking we'd then potentially be modifying the cached data (when it's dirty) in multiple places concurrently. Which night lead to some unexpected results.

mopidy_spotify/playlists.py
```diff
@@ -93,42 +59,30 @@ def save(self, playlist):
...
+        bitrate=bitrate,
+        as_items=as_items,
     )
 
-    # XXX Should Mopidy support more fine grained playlist events which this
-    # event can trigger?
-
 
-def on_playlist_moved(sp_playlist_container, sp_playlist, old_index, new_index):
+def on_playlists_loaded():
```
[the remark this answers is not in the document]
kingosticks: Hmm I think I had intended to call it from `refresh`. It feels as if we should be calling it somewhere.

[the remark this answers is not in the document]
kingosticks: I guess I'll get rid of it for now and we add it back if/when we sort out the playlist modification functionality down the line.

kingosticks:
I did find a bug where the first page of a web object won't update the etag. That bug wasn't introduced in this PR so it's not a blocker to merging this now.

mopidy_spotify/web.py
```diff
@@ -353,3 +353,27 @@ def __str__(self):
             f"URL: {self.url} ETag: {self._etag} "
             f"Expires: {datetime.fromtimestamp(self._expires)}"
         )
+
+
+class SpotifyOAuthClient(OAuthClient):
+    def __init__(self, client_id, client_secret, proxy_config):
```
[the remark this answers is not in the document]
jodal: The `*` forces all callers of the method to pass all arguments after the star as kwargs. That makes APIs safer for future extension, as you know that nothing can depend on the order of arguments after the `*`.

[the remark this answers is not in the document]
kingosticks: Ahh OK that sounds good.

mopidy_spotify/playlists.py
jodal: Yeah, we're in an actor here, so we have by definition only one request to process at a time. Thus, if you set `self._loaded = False` at the beginning of the method should have no effect, as no other code will run and be able to read it before you set it to `True` at the end of the method.

mopidy_spotify/playlists.py
```diff
@@ -1,90 +1,74 @@
...
+            self._backend._bitrate,
+            as_items,
         )
 
     def refresh(self):
-        pass  # Not needed as long as we don't cache anything.
+        if not self._backend._web_client.logged_in:
+            return
+
+        with utils.time_logger("playlists.refresh()", logging.INFO):
```
jodal: Should we reduce this to `DEBUG` level?

mopidy_spotify/playlists.py
```diff
@@ -1,90 +1,74 @@
...
             playlist_ref = translator.to_playlist_ref(
-                sp_playlist, folders=folders, username=username
+                web_playlist, web_client.user_id
             )
             if playlist_ref is not None:
                 yield playlist_ref
 
     def get_items(self, uri):
-        with utils.time_logger(f"playlist.get_items({uri})"):
+        with utils.time_logger(f"playlist.get_items({uri!r})", logging.INFO):
```
jodal: And this to `DEBUG` level?

mopidy_spotify/playlists.py
```diff
@@ -1,90 +1,74 @@
...
 
 class SpotifyPlaylistsProvider(backend.PlaylistsProvider):
     def __init__(self, backend):
         self._backend = backend
         self._timeout = self._backend._config["spotify"]["timeout"]
+        self._loaded = False
 
     def as_list(self):
-        with utils.time_logger("playlists.as_list()"):
+        with utils.time_logger("playlists.as_list()", logging.INFO):
```
jodal: Reduce to `DEBUG` level?

jodal:
And a changelog entry!

kingosticks:
Yes to reducing the log levels. I'll also change `get_one` to use trace logging.

kingosticks:
Done and done.

jodal:
:tada: :rocket:

BlackLight:
Is Python 2 still supported? I've tried to run the installation after merging my changes for playlist modifications, and I've got `setup.py` failing on Python 3 format strings:

```
SyntaxError: invalid syntax
    logger.info(f'Failed to browse "{uri}": Toplist URI parsing failed')
```

If Python 2 is no longer supported does it mean that the master for this extension (and for mopidy itself and for most of its extensions) is now supposed to be run only on Python 3?

Don't get me wrong, it's about time that mopidy finally moves to Python 3, but there's tons of extensions out there that also need to be migrated :)

kingosticks:
> Is Python 2 still supported?

The next planned release of Mopidy-Spotify is v4.0.0 and it will [require Mopidy v3.0.0+](https://github.com/mopidy/mopidy-spotify/blob/master/setup.cfg#L28). Python 2 will **not** be supported in these releases.

> does it mean that the master for this extension (and for mopidy itself and for most of its extensions) is now supposed to be run only on Python 3?

Mopidy's `master` branch still reflects the latest release and thus supports _Python 2 only_. Mopidy's `develop` branch contains v3.x work and supports _Python 3 only_. Extensions (at least those living under mopidy-org) now only have a `master` branch and soon they will all support _Python 3 only_. We decided that having a `develop` branch alongside `master` isn't worth the extra complication (particuarly during releases) because for extensions they are almost identical most of the time. We are currently in an unsual period of enormous change so `master` has diverged signicatly from what was last released - hopefully it won't last much longer. 

> but there's tons of extensions out there that also need to be migrated :)

We are very aware! And we've been working on that: https://github.com/orgs/mopidy/projects/2

If you have more playlist enhancements that can go into Mopidy v4 we'd love to have them.

jodal:
We've also built out a new extension registry to make it easy for end-users to find extensions that will work with Python 3: https://mopidy.com/ext/

What's lacking is outreach to the long tail of extension developers. I think that should be our first focus after the Mopidy 3.0 release, which is targeted at Dec 21.

PR #250, opened 2019-12-29
Initial support for saving Spotify playlists.

fatg3erman:
This adds support for the 'save' function in playlists.py, using the web api client.

Pretty simple stuff, just had to add an extra parameter so I could force it to refresh the playlist after the save, otherwise it returned the cached copy which made it appear to a mopidy client that nothing had happened.

### DIFF
--- a/mopidy_spotify/browse.py
+++ b/mopidy_spotify/browse.py
@@ -148,7 +148,7 @@ def _browse_toplist_user(web_client, variant):
         return []
 
     if variant in ("tracks", "artists"):
-        items = web_client.get_one(f"me/top/{variant}").get("items", [])
+        items = web_client.get_one(f"me/top/{variant}", True).get("items", [])
         if variant == "tracks":
             return list(
                 translator.web_to_track_refs(items, check_playable=False)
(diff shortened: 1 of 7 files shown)


fatg3erman:
OK I've had enough now. I don't understand why 'black' is failing and it won't tell me.

If someone can tell me if you're interested in merging this and explain why it's failing, I'll finish it.

kingosticks:
Is this radical different to the other or here that aims to do the same thing?

fatg3erman:
Ha! Sorry I didn't even notice the other one!

This attempts to do the same thing - in fact the other one does it better by atomizing the changes. However, the other one clears the cache to work around the caching issue, whereas this one adds a new use_cache parameter to get_one() which allows us to refresh just the one playlist while keeping the rest of the cache intact. That said, on thinking about it, I think that might not work because the old version will still be in the cache next time it gets looked up.

Currently suffering with insomnia. This is the result of early-morning boredom :) Perhaps best to ignore this.

fatg3erman:
Yeah, this is actually not very good. Sorry for the bother.

PR #269, opened 2020-07-24
Fix gapless playback

### LINKED ISSUE #160: Add gapless playback
Especially when playing back live-albums on Spotify I can recognize small pauses between the tracks. Would be great, if that issue could be fixed.

From my experience MPD has the ability for gapless playback (buffering). Don't know if this is an issue for mopidy or mopidy-spotify.

Soaku:
After the song ends, libspotify continues to push data to the buffer, which breaks gapless playback.

This change fixes this by unloading the track after it ends. 

Closes #160.

### DIFF
--- a/mopidy_spotify/playback.py
+++ b/mopidy_spotify/playback.py
@@ -15,6 +15,9 @@ GST_CAPS = "audio/x-raw,format=S16LE,rate=44100,channels=2,layout=interleaved"
 # Extra log level with lower importance than DEBUG=10 for noisy debug logging
 TRACE_LOG_LEVEL = 5
 
+# Last audio data sent to the buffer, currently on hold
+_held_buffer = None
+
 
 class SpotifyPlaybackProvider(backend.PlaybackProvider):
     def __init__(self, *args, **kwargs):
@@ -50,6 +53,8 @@ class SpotifyPlaybackProvider(backend.PlaybackProvider):
     def change_track(self, track):
         self._connect_events()
 
+        global _held_buffer
+
         if track.uri is None:
             return False
 
@@ -78,6 +83,7 @@ class SpotifyPlaybackProvider(backend.PlaybackProvider):
             sp_track.load(self._timeout)
             self.backend._session.player.load(sp_track)
             self.backend._session.player.play()
+            _held_buffer = None
 
             future = self.audio.set_appsrc(
                 GST_CAPS,
@@ -162,6 +168,8 @@ def music_delivery_callback(
     # This is called from an internal libspotify thread.
     # Ideally, nothing here should block.
 
+    global _held_buffer
+
     if seeking_event.is_set():
         # A seek has happened, but libspotify hasn't confirmed yet, so
         # we're dropping all audio data from libspotify.
@@ -187,8 +195,16 @@ def music_delivery_callback(
         bytes(frames), timestamp=buffer_timestamp.get(), duration=duration
     )
 
+    if _held_buffer:
+
+        # Send any queued buffer first.
+        future = audio_actor.emit_data(_held_buffer)
+
+    # libspotify sends an empty buffer before calling end_of_track_callback, so we need to hold it.
+    _held_buffer = buffer_
+
     # We must block here to know if the buffer was consumed successfully.
-    consumed = audio_actor.emit_data(buffer_).get()
+    consumed = future.get()
 
     if consumed:
         buffer_timestamp.increase(duration)
@@ -208,7 +224,7 @@ def end_of_track_callback(session, end_of_track_event, audio_actor):
     end_of_track_event.set()
     audio_actor.emit_data(None)
 
-    # unload the player to stop receiving data
+    # Unload the player to stop receiving data
     session.player.unload()
 
 
--- a/mopidy_spotify/playback.py
+++ b/mopidy_spotify/playback.py
@@ -208,6 +208,9 @@ def end_of_track_callback(session, end_of_track_event, audio_actor):
     end_of_track_event.set()
     audio_actor.emit_data(None)
 
+    # unload the player to stop receiving data
+    session.player.unload()
+
 
 class BufferTimestamp:
     """Wrapper around an int to serialize access by multiple threads.
--- a/mopidy_spotify/playback.py
+++ b/mopidy_spotify/playback.py
@@ -195,16 +195,14 @@ def music_delivery_callback(
         bytes(frames), timestamp=buffer_timestamp.get(), duration=duration
     )
 
+    # If there is any held buffer, send it
     if _held_buffer:
-
-        # Send any queued buffer first.
-        future = audio_actor.emit_data(_held_buffer)
+        consumed = audio_actor.emit_data(_held_buffer).get()
+    else:
+        consumed = True
 
     # libspotify sends an empty buffer before calling end_of_track_callback, so we need to hold it.
     _held_buffer = buffer_
-
-    # We must block here to know if the buffer was consumed successfully.
-    consumed = future.get()
 
     if consumed:
         buffer_timestamp.increase(duration)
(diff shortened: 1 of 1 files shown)


Soaku:
Weird, this fix doesn't seem to be working when running from `systemctl`. I wonder why, logger calls I placed in the code are added to the log, so it must be running with my patch added, but the pause between songs is still there.

Soaku:
Ok, apparently libspotify sends one of those invalid packets before sending track end signal... is there any way to stop that other than measuring buffer size?

adamcik:
This has been eluding me for years, this is an awesome find! Perhaps you could delay everything by one buffer by queuing it in our code (not in GStreamer) and then just discard the last one if this bad behavior is consistent?

Soaku:
@adamcik Thank you! I thought of this solution, and I think it should work well.

I'll ask first, the callbacks do not have access to `self`, do they? If I were to implement the delay, where should I put the variable? I'm used to avoiding globals, so is a module-scoped variable okay?

adamcik:
We know this should be a singleton, so I think we can get away with doing this module scoped. At least it's good enough to get things going and see if it works?

Soaku:
Sure, it's working. Ready for review.

Soaku:
Sorry for that commit, I'm stupid.

Soaku:
Too many commits for this little thing, excuse my stupidity. At least it's working well now, I've been playing music and radios all day, I'm not getting any errors and transitions are good.

@adamcik I'd like to request a review.

mopidy_spotify/playback.py
```diff
@@ -15,6 +15,9 @@
 # Extra log level with lower importance than DEBUG=10 for noisy debug logging
 TRACE_LOG_LEVEL = 5
 
+# Last audio data sent to the buffer, currently on hold
```
adamcik, line 18: I would put in a small explanation about why we need to do this, it's not obvious without the context of this PR. So perhaps a short sentence with a link to this PR or a self contained paragraph?

mopidy_spotify/playback.py
```diff
@@ -78,6 +83,7 @@ def change_track(self, track):
             sp_track.load(self._timeout)
             self.backend._session.player.load(sp_track)
             self.backend._session.player.play()
+            _held_buffer = None
```
adamcik, line 86: @jodal probably knows the answer to this better than me, but when would the session start being able to fill this value async? I.e. should we reset it before calling load/play, or does it depend on the current state of the session?

Could protecting this global with a lock be reasonable, or are we just asking for deadlocks if we aren't careful?

mopidy_spotify/playback.py
```diff
@@ -187,8 +195,14 @@ def music_delivery_callback(
         bytes(frames), timestamp=buffer_timestamp.get(), duration=duration
     )
 
-    # We must block here to know if the buffer was consumed successfully.
-    consumed = audio_actor.emit_data(buffer_).get()
+    # If there is any held buffer, send it
+    if _held_buffer:
+        consumed = audio_actor.emit_data(_held_buffer).get()
+    else:
+        consumed = True
```
adamcik, line 202: I don't know if this should be true when there isn't a held buffer, as a follow on effect would be increasing the duration incorrectly?

adamcik:
I've added some suggestions, and mostly questions about things I don't know the answer to off the top of my head. It's been quite some time since I've been deep in the audio code for mopidy, at some point my attempts at getting gapless working and unblocking plugable outputs ground to a halt likely due to what this PR fixes.

adamcik:
Also, if you feel your git skills are up to it, feel free to tidy up the commits by rebasing/squashing etc (but no pressure if you feel that would just create a mess). We often do this type of cleanup on our own changes to keep the history tidy. As we'll often go through a similar process as you have in this PR finding more things to fix/adjust...

adamcik:
@jodal can/should the empty buffer suppression be moved to pyspotify, or should we document change things so we could do a `buffer == pyspotify.SOME_CONSTANT_FOR_THE_FINAL_SAMPLE` type check akin to checking for EOS in the GStreamer world?

mopidy_spotify/playback.py
Soaku: I set it as true, because we take the current buffer and put it on hold, which (if my understanding of frame consumption is right) means we did consume the frames — even if we'll push it to Mopidy later. 

On the other hand, I'm not sure if the `buffer_timestamp` is increased correctly if its done before emitting data.

kingosticks:
Nice! This is a great find and I don't understand how I missed / forgot about this PR. It prompted me to find https://stackoverflow.com/questions/26014520/libspotify-c-sending-zeros-at-the-end-of-track which also talks about that final data buffer of 22050 zeros. I think your solution works great and I am happy to go with it. Catching this funny final music_delivery case as it happens is also an option. We could then avoid emitting the large empty buffer which is the ultimate cause of our nasty gap. Something like:
```
    if num_frames == pyspotify.NUM_FRAMES_EOT: # 22050
        consumed = True
      else:
        consumed = audio_actor.emit_data(buffer_).get() 
```
The above works with or without the addition of `session.player.unload()` but we might as well have that too since it makes sense. 

That Stack Overflow post also talks about how the duration is pushed beyond the length of the track by this fake final delivery and otherwise may come up a little short. So another option might be detect when the `buffer_timestamp` increase is going to take us beyond the length of the track.

Soaku:
@kingosticks Your proposed solution is what I thought of at first and used as a workaround, but I wasn't sure if this condition would always trigger only when needed. I think buffer holding is better for this reason.

Also, sorry for stalling the PR for the time, I hope the added comments should be okay now.

By the way, because I was away and my laptop had weird issues with Python, I had to use plain Spotify for a week. Their clients are so inconsistent and broken. Mopidy is really a bless. Thank you for making this project :heart:

mopidy_spotify/playback.py
kingosticks: It does seem weird to attach what we know is the *wrong* timestamp to the buffer we are emitting to GST (although the error will be very small). Should we be doing something more like `buffer_timestamp.increase(_held_buffer.duration)`?

kingosticks:
I understand and I do agree that delaying the buffer is nicer in theory, but involving inter-thread communication makes it more complicated too.

mopidy_spotify/playback.py
kingosticks: I think this is fine. It doesn't matter when we increment `buffer_timestamp` as long as it's before we use in the next delivery.

kingosticks:
And I will add, I love that you've fixed this. There're some albums I've only ever heard on Spotify and they sound almost brand new (better!) with gapless fixed - amazing!

mopidy_spotify/playback.py
Soaku: It should be fine if placed before. My mistake I didn't think it could happen. I'll correct this.

jodal (approved):
I think this looks good! All the testing and eyes on what is quite small changes make me confident this is a clear improvement.

I prefer holding the buffer to discarding deliveries of exactly 22050 frames. 22050*2=44100, as in https://en.wikipedia.org/wiki/44,100_Hz, so that sounds like exactly 0.5s of audio and not an obscure sample size we'll never see in other situations.

jodal:
Regarding the failing CI checks:

- If you rebase one of the three flake8 warnings will be fixed. The other two are comment lines from this PR that needs to be wrapped to be shorter.
- The single test that fails, 'test_music_delivery_creates_gstreamer_buffer_and_gives_it_to_audio', I think just needs to deliver two buffers instead of one, and still assert that just a single buffer is delivered to Mopidy's audio subsystem.

kingosticks:
What happens with the global `_held_buffer` under pytest? Don't they all run in parallel and all access the same variable?

kingosticks:
In fact, and even when not running in parallel, `_held_buffer` isn't reset to `None` between all those `music_delivery_callback` tests.

kingosticks:
Closing this in favour of the version at #277 that's tweaked for easier testing.

PR #270, opened 2020-08-18
Update README.rst

celowsky:
Includes additional dependencies required by an install with a Raspberry Pi 3B+


jodal:
These are not Mopidy-Spotify's direct dependencies, but transitive dependencies. If you're using a package manager, it is handled for you. If you install from source, these are already documented by the project using these, pyspotify.

PR #273, opened 2020-08-25
Stop using deprecated yield_fixture

### LINKED ISSUE #251: Stop using deprecated yield_fixture
> Since pytest-3.0, fixtures using the normal fixture decorator can use a yield statement to provide fixture values and execute teardown code, exactly like yield_fixture in previous versions.
>
>Marking functions as yield_fixture is still supported, but deprecated and should not be used in new code.

https://docs.pytest.org/en/latest/yieldfixture.html

L3viathan:
Closes #251.


jodal (approved):
Thanks! :-)

PR #277, opened 2020-09-29
Soaku gapless playback

kingosticks:
This is #269 with the remaining CI fixes. And also with the held buffer changed from a global to an instance variable as the `music_delivery_callback` tests were no longer doing the right thing.

### DIFF
--- a/tests/test_playback.py
+++ b/tests/test_playback.py
@@ -64,6 +64,7 @@ def test_connect_events_adds_music_delivery_handler_to_session(
             playback_provider._seeking_event,
             playback_provider._push_audio_data_event,
             playback_provider._buffer_timestamp,
+            playback_provider._held_buffer,
         )
         in session_mock.on.call_args_list
     )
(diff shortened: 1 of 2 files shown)


kingosticks:
There is still https://github.com/mopidy/mopidy-spotify/pull/277/files#diff-d09e8ce8f0a316832307028fe46a8319R202 to think about. I think it's actually OK and the None condition is handled by our appsrc.

Soaku:
Haven't noticed your PR before. I'm glad you fixed it and I think your solution should be okay.

tests/test_playback.py
```diff
@@ -388,6 +449,7 @@ def test_music_delivery_consumes_zero_frames_if_audio_fails(
     push_audio_data_event.set()
     buffer_timestamp = mock.Mock()
     buffer_timestamp.get.return_value = mock.sentinel.timestamp
+    held_buffer = playback.BufferData(mock.sentinel.gst_buffer)
```
jodal, line 452:
```suggestion
    held_buffer = playback.BufferData(mock.sentinel.gst_buffer1)
```

jodal (approved):
:tada:

jodal:
After an hour or so on video with @kingosticks live-coding now, we've:

- Made it thread-safe, in case someone changes track in the middle of doing an operation on the held buffer.
- Made sure to hold on to the right buffer if the appsrc applies backpressure.

Soaku:
This is great news! Thank you!

PR #279, opened 2020-10-04
Load all tracks in "Your Music", "Playlists" and "Top Lists"

### LINKED ISSUE #256: "Your Music" limited to 20 results
When browsing "Your Music" categories ("Your Tracks", "Your Albums") only 20 results are being returned. Initially, I thought that I just needed to set the `set_[category]_count` value in the config file, but changing this value has no impact on the results being returned (it does seem to work fine for 'normal' searches, but doesn't change the number of results for "Your [Category]").

Since 20 is the default limit for the relevant requests in Spotify's Web API ([here](https://developer.spotify.com/documentation/web-api/reference-beta/#endpoint-get-users-saved-albums)), I'm wondering if the `limit` param is even being used.

Anyway, is this intended? I know most people just make playlists but I use my saved albums pretty heavily and it was a bit jarring to find that they don't all get loaded.

girst:
fixes #256

### DIFF
--- a/mopidy_spotify/browse.py
+++ b/mopidy_spotify/browse.py
@@ -4,6 +4,7 @@ from mopidy import models
 
 import spotify
 from mopidy_spotify import countries, playlists, translator
+from mopidy_spotify.utils import flatten
 
 logger = logging.getLogger(__name__)
 
@@ -148,7 +149,16 @@ def _browse_toplist_user(web_client, variant):
         return []
 
     if variant in ("tracks", "artists"):
-        items = web_client.get_one(f"me/top/{variant}").get("items", [])
+        items = flatten(
+            [
+                page.get("items", [])
+                for page in web_client.get_all(
+                    f"me/top/{variant}",
+                    params={"limit": 50},
+                )
+                if page
+            ]
+        )
         if variant == "tracks":
             return list(
                 translator.web_to_track_refs(items, check_playable=False)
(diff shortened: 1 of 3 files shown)


kingosticks (approved):
It's close to being worth writing a helper to flatten these pages of results. It would be nice if the mocked results could return multiple pages (including maybe some empty ones) to better test the new code. But it does of course all work just fine. I did find it a little slow for my larger "your music" list to load and I would guess that's going to be very noticeable for people on slow machines. Some "[repagination](https://github.com/mopidy/mopidy-dirble/pull/19/files#diff-54cc99286f7e9dade0b2e7d8874bf90fR22)" would be nice but we can solve that another day.

girst:
On Mon, Nov 02, 2020 at 10:12:35AM -0800, Nick Steel wrote:
>@kingosticks approved this pull request.
>
>It's close to being worth writing a helper to flatten these pages of 
>results.

agreed. will do.

>It would be nice if the mocked results could return multiple pages 
>(including maybe some empty ones) to better test the new code. But it 
>does of course all work just fine.

i'll try, but no promises.

>I did find it a little slow for my larger "your music" list to load and 
>I would guess that's going to be very noticeable for people on slow 
>machines. Some "[repagination]" would be nice but we can solve that 
>another day.

I think i understand what you mean, but can't comprehend the patch 
you've linked.

girst:
@kingosticks: implemented a flatten() helper and modified tests to check 
for two pages. i believe this is ready for merge now.

thanks for the review!

kingosticks (approved):

kingosticks:
This is great - thank you.

PR #287, opened 2020-11-13
Bug22: Support for playlist management (v2)

### LINKED ISSUE #22: Add support for Spotify playlist management

girst:
This is now 95% there, and I'd like to ask for some feedback now.

I've adapted Myers' diffing algorithm (the one that's used by e.g. git-diff) to support moving ranges of tracks. It generates a relatively optimal edit script, to minimize API requests without the diffing algorithm taking too long. Especially getting moves to work was quite tricky (took me over a week of daily head-scratching), but I have written a whole bunch of tests for it. the web_client_mock now also supports editing playlists.

One thing we haven't yet talked about is what to do when API requests fail. The worst case (and only serious one) is when we clear-and-repopulate a playlist. When we fail mid-way through, we lose data. Should we do something (e.g. save an m3u playlist somewhere) then?

Also, there's still a bug with remove_from_cache() right now (when adding tracks the view is not updated, only when deleting tracks do all previous changes then show up. this also causes problems with moving tracks). I guess I am not invalidating everything I should, as it works when reverting to clear_cache (pointers appreciated).

What's still todo is:
- [ ] thoroughly testing it out with the actual spotify api
- [ ] figure out what to do on an API error
- [ ] extending the authenticator on the website to request additional scopes
- [ ] maybe write some more tests (see the comment in test_playlists.py)
- formatting everything with black and such

CC: @BlackLight
closes #22, closes #236


mopidy_spotify/playlists.py
```diff
@@ -49,13 +53,94 @@ def _get_playlist(self, uri, as_items=False):
...
+    @staticmethod
+    def _span(p, l):  # like haskell's Data.List.span
+        i = next((i for i,v in enumerate(l) if not p(v)), len(l))
+        return l[:i], l[i:]
+
+    def _patch_playlist(self, playlist, operations):
+        delta_f = 0
+        delta_t = 0
+        unwind_f = []
+        unwind_t = []
```
girst, line 76: I'm a bit annoyed about having two of each here, but for moving tracks around, there are stupid edge cases otherwise :/

mopidy_spotify/translator.py
```diff
@@ -315,7 +315,10 @@ def web_to_album(web_album):
     ]
     artists = [a for a in artists if a]
 
-    return models.Album(uri=ref.uri, name=ref.name, artists=artists)
+    # Note: date can by YYYY-MM-DD, YYYY-MM or YYYY.
+    date = web_album.get('release_date', '').split('-')[0] or None
+
+    return models.Album(uri=ref.uri, name=ref.name, artists=artists, date=date)
```
girst, line 321: I had to extend the track translator to parse release dates (just the bare minimum (year); spotify also supports finer dates) to support round-tripping between the web and native formats. otherwise, I wouldn't have been able to compare output from save() and lookup().

girst:
I found and fixed the problem with the cache (the playlist's tracks are fetched from two endpoints, not just the one I cleared). While at it, I also moved the newly added code to use the new shorter playlist URIs that spotify offers (as the old ones are deprecated). Also started testing this out on my personal spotify account, and it pretty much works.

This leaves us with only one major item to figure out: what do do on errors. For this, I'd like to solicit some ideas from you guys.

girst:
OK, so I think this is done. I'm going to add @jodal, @adamcik and @kingosticks as potential reviewers. Looking forward to your feedback!

kingosticks (changes requested):
I came across another general issue. We currently filter out all unplayable tracks and this means we'll be removing them whenever we make any playlist change, which is wrong. Tracks might be currently unplayable if unavailable in the geo region where the user is logged in (but may be playable from elsewhere and/or in the future). I think local files might also be treated as unplayable but I'm not sure if Spotify still supports this feature I never used it. I think to workaround this we'll need to change the current playlist behaviour (which has been the same since our first implementation) to now include unplayable tracks in playlists. libspotify won't be able to load these tracks when it comes to try and play them and they will be skipped with an `info` log message (which we will want to adjust). These unplayable tracks will clutter the queue/playlist listing and might be confusing to users when they try to play them. But I don't see another way to avoid losing them when we use `save()`.

I've not looked closely at the myers implementation here, I assume it's the normal one that's available everywhere. It seems odd that in 2020 there is no standard, fast (C), robust implementation available for us to just pull in and use. It's not ideal we have to reproduce (and test) our own version of this common algorithm. But we do have Python's standard `difflib.SequenceMatcher`, so what does using myers provide over just using that? I did find multiple (why???) packages for Google's [diff-match-patch(https://github.com/google/diff-match-patch) library. That seems to use many of the optimisations we originally talked about in #236 but I've no great interest in using that, seems overly complicated. 

At a higher level, some of what has been added to playlist.py belongs in web.py. You'll notice there isn't currently any Spotify Web API specific stuff in playlist.py. I think we want to keep that separation.

mopidy_spotify/utils.py
```diff
@@ -33,3 +35,113 @@ def time_logger(name, level=TRACE):
 
 def flatten(list_of_lists):
     return [item for sublist in list_of_lists for item in sublist]
+
+
+class op:
+    def __init__(self, op, tracks, frm, to=0):
+        self.op = op
+        self.tracks = tracks
+        self.frm = frm
```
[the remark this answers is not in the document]
girst: `from` is a reserved keyword, so it can't be used

mopidy_spotify/utils.py
```diff
@@ -33,3 +35,113 @@ def time_logger(name, level=TRACE):
...
+        self.to = to
+
+    def __repr__(self):
+        length = len(self.tracks)
+        first = self.tracks[0].split(":")[-1]
```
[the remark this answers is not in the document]
girst: this is just helpful for debugging; it's not used anywhere (except for printf-debugging). regarding the possibility of an index error, that won't happen, since we never construct op()s with no tracks inside.

if you want, i can remove it completely.

mopidy_spotify/utils.py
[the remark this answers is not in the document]
kingosticks: Argh, ok, fine. I just have bad memories of this practice (the Perl days).

mopidy_spotify/utils.py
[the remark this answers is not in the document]
kingosticks: If we leave it in, someone will inevitably find a way of calling it without that condition in the future. So we either fix it or remove it. If the debug output is still useful then please fix it.

kingosticks:
Link to the [Zulip discussion](https://mopidy.zulipchat.com/#narrow/stream/207265-mopidy-dev/topic/.5BPULL.5D.20mopidy-spotify.3A.20playlist.20editing)

girst:
> We currently filter out all unplayable tracks and this means we'll be removing them whenever we make any playlist change, which is wrong.

agreed, that's bad.

> I think local files might also be treated as unplayable but I'm not sure if Spotify still supports this feature I never used it.

i don't have any experience here either. looking at [the docs](https://developer.spotify.com/documentation/general/guides/local-files-spotify-playlists/), uris look like `spotify:local:{artist}:{album_title}:{track_title}:{duration_in_seconds}`, so that's something we could simulate.

> change the current playlist behaviour  to now include unplayable tracks in playlists.
> info log message (which we will want to adjust).

that seems like a good plan. should we do this in a followup to keep this patch from growing even further?

> I've not looked closely at the myers implementation here, I assume it's the normal one that's available everywhere.

yep, pretty much. it has a extra memory optimisation, but that's about it.

> But we do have Python's standard difflib.SequenceMatcher, so what does using myers provide over just using that?

I was under the impression that difflib only worked on strings of text, not lists as I used here. Your comment made me look again, and it turns out I was wrong. That said, the interface is pretty different to my version. I'm up for rewriting it, but it'll take me a few days. I think we can continue the review with the current implementation, and have a drop-in replacement for myers() and diff() later on.

> I did find multiple (why???) packages for Google's diff-match-patch library.

I'd rather avoid introducing 3rd party dependencies if possible. 

> At a higher level, some of what has been added to playlist.py belongs in web.py.

I guess, you#re talking about _patch_playlist(), _replace_playlist() and _playlist_edit()? i can move that functionality to web.py.

girst:
ok, myers diff is gone. porting to difflib turned out easier than expected, once i familiarized myself with it.

mopidy_spotify/utils.py
```diff
@@ -33,3 +35,71 @@ def time_logger(name, level=TRACE):
...
+
+    return list(partition(o, chunksize)) if o.op != "m" else [o]
+
+
+def diff(old, new, chunksize=100):
+    # first, apply python's built-in diff algorithm, remove unmodified ranges,
+    # split replacements into seperate insertions and deletions and transform
+    # the data structure into op-objects:
+    ops = []
+    for tag, i1, i2, j1, j2 in difflib.SequenceMatcher(
```
kingosticks, line 79: I think we also want `autojunk=False`.

girst:
OK, coming back to the "not filter out unplayable/local tracks" issue: I've pushed a commit, 6667e0e, that should "fix" that. I haven't tested that, as I don't have any unplayable tracks in my playlists AFAIK. Maybe you can have a look at that.

girst:
@kingosticks, sorry for the delay, but `spotify:local` is now handled (as in "refused"). This should have all your (valuable, btw) input addressed now, no?

mopidy_spotify/translator.py
```diff
@@ -315,11 +322,14 @@ def web_to_album(web_album):
     ]
     artists = [a for a in artists if a]
 
-    return models.Album(uri=ref.uri, name=ref.name, artists=artists)
+    # Note: date can by YYYY-MM-DD, YYYY-MM or YYYY.
+    date = web_album.get("release_date", "").split("-")[0] or None
```
kingosticks, line 326: Turns out I do have some local tracks in one of my playlists! And they are annoying to handle because all the fields are present, but most are None. 
```
SpotifyBackend-4 ERROR    2020-12-04 10:19:15,616 Unhandled exception in SpotifyBackend (urn:uuid:2a48d2d3-a581-489a-b0ed-1a3ad4d321ba):
Traceback (most recent call last):
  File "/home/nick/.virtualenvs/mopidy3/lib/python3.8/site-packages/pykka/_actor.py", line 186, in _actor_loop
    self.on_start()
  File "/home/nick/Dev/mopidy-dev/mopidy-spotify/mopidy_spotify/backend.py", line 68, in on_start
    self.playlists.refresh()
  File "/home/nick/Dev/mopidy-dev/mopidy-spotify/mopidy_spotify/playlists.py", line 140, in refresh
    self._get_playlist(playlist_ref.uri)
  File "/home/nick/Dev/mopidy-dev/mopidy-spotify/mopidy_spotify/playlists.py", line 49, in _get_playlist
    return playlist_lookup(
  File "/home/nick/Dev/mopidy-dev/mopidy-spotify/mopidy_spotify/playlists.py", line 265, in playlist_lookup
    playlist = translator.to_playlist(
  File "/home/nick/Dev/mopidy-dev/mopidy-spotify/mopidy_spotify/translator.py", line 240, in to_playlist
    tracks = [t for t in tracks if t]
  File "/home/nick/Dev/mopidy-dev/mopidy-spotify/mopidy_spotify/translator.py", line 240, in <listcomp>
    tracks = [t for t in tracks if t]
  File "/home/nick/Dev/mopidy-dev/mopidy-spotify/mopidy_spotify/translator.py", line 233, in <genexpr>
    web_to_track(
  File "/home/nick/Dev/mopidy-dev/mopidy-spotify/mopidy_spotify/translator.py", line 341, in web_to_track
    album = web_to_album(web_track.get("album", {}))
  File "/home/nick/Dev/mopidy-dev/mopidy-spotify/mopidy_spotify/translator.py", line 326, in web_to_album
    date = web_album.get("release_date", "").split("-")[0] or None
AttributeError: 'NoneType' object has no attribute 'split'
```
Caused by the following web_track:
```
{'album': {'album_type': None, 'artists': [], 'available_markets': [], 'external_urls': {}, 'href': None, 'id': None, 'images': [], 'name': 'Never Mind The Bollocks [Heres The Sex Pistols]', 'release_date': None, 'release_date_precision': None, 'type': 'album', 'uri': None}, 'artists': [{'external_urls': {}, 'href': None, 'id': None, 'name': 'Sex Pistols', 'type': 'artist', 'uri': None}], 'available_markets': [], 'disc_number': 0, 'duration_ms': 211000, 'explicit': False, 'external_ids': {}, 'external_urls': {}, 'href': None, 'id': None, 'is_local': True, 'name': 'Anarchy In The U.K.', 'popularity': 0, 'preview_url': None, 'track_number': 0, 'type': 'track', 'uri': 'spotify:local:Sex+Pistols:Never+Mind+The+Bollocks+%5BHeres+The+Sex+Pistols%5D:Anarchy+In+The+U.K.:211'}
```
This issue has always been there but we just never bumped into it. I submitted https://github.com/mopidy/mopidy-spotify/pull/290 to deal with it since it's really a more general bug.

girst:
Urgh, I _so_ wish, python would have had null-coalescing/none-aware 
operators much earlier, then the whole flagile .get()-dance could be 
avoided alltogether.

Thanks for finding this; I'll pull from master once your PR is merged.

girst:
Oh, and if you're feeling adventurous, @kingosticks, i'd be very 
interested in what happens when you patch out the last commit (refuse 
with local tracks present) and try to delete a local track (probably 
wise to do in a dedicated test playlist. note that the web api can't add 
local tracks, so you'd have to do that outside mopidy).  the docs say 
local tracks "should" be deleted using different arguments to the delete 
call, but I wonder what would happen when ignoring this advice.

If you don't have the time, no worries, though. If it worked at all, I 
would if at all implement local track deletion in a seperate PR, not 
that one.

kingosticks:
Have merged #290 

I'm not sure if the playlist that was causing this was actually mine or just one I follow. But I will have a little play if I get more time. I am going to prioritise a more thorough review of this as it is.

girst:
On Sat, Dec 05, 2020 at 04:51:07PM -0800, Nick Steel wrote:
>Have merged #290

Pulled.

>I am going to prioritise a more thorough review of this as it is.

Appreciated, thanks!

girst:
I hate to be pushy, but do you think you can find some time over the 
holidays to review this, @kingosticks? Let's end 2020 on a high note 
(pun intended) ;-)

kingosticks:
I know! This has been on my todo list for so long, I apologise. Had a busy end to the year but holidays have started now so is time before Christmas. Hopefully we can rope some others in too.

BlackLight:
I've tested the branch and taken a look at the code and it all seems good to me - can't wait for this feature to be finally merged.

mopidy_spotify/translator.py
```diff
@@ -214,6 +214,7 @@ def to_playlist(
     bitrate=None,
     as_ref=False,
     as_items=False,
+    check_playable=True,
```
kingosticks, line 217: The functionality of this new parameter is not tested.

girst: check_playable is passed straight through to web_to_track_refs, which is tested in https://github.com/mopidy/mopidy-spotify/blob/master/tests/test_translator.py#L454

mopidy_spotify/translator.py
kingosticks, line 326: `TestWebToAlbum` should also include a test for this.

tests/test_web.py
```diff
@@ -663,6 +663,24 @@ def test_cached_response_unchanged(web_response_mock, oauth_client, mock_time):
...
+    responses.add(responses.DELETE, "https://api.spotify.com/v1/foo", json={})
+    oauth_client._expires = 2000
+    mock_time.return_value = 1
+
+    result = oauth_client.post("foo")
+    result = oauth_client.put("foo")
+    result = oauth_client.delete("foo")
+
+    assert len(responses.calls) == 3
+    assert not result.status_unchanged
```
kingosticks, line 681: This only tests the final result you got from `delete(). And what is this trying to test? It also passes if the last one is changed to `result = oauth_client.get("foo")`.

kingosticks (changes requested):
I have not reviewed the playlists tests because
1. This review is already lagging my browser
2. I think this PR is already quite big to the point where there's benefit in someone else writing the tests for it. We normally do not go for the style of mocking out the API functionality like this. I understand it results in less boilerplate/repetition in the actual tests but it's more complex and I think we should stick to the Responses-based approach we've used so far.

I am also curious as to why the playlist tests run so slowly. 

EDIT:

> where there's benefit in someone else writing the tests for it.

And to be clear, there's a lot of great stuff in there and I really appreciate the time and effort that went into it. I think most of it can be used as is, there's just some bits we could do differently. I will push some commits to your branch (I think that is probably the best way) over the next couple of days.

mopidy_spotify/playlists.py
```diff
@@ -65,16 +144,114 @@ def refresh(self):
...
+        saved_playlist = self._get_playlist(playlist.uri, with_owner=True)
+        if not saved_playlist:
+            return
+
+        saved_playlist, owner = saved_playlist
+        # We limit playlist editing to the user's own playlists, since mopidy
+        # mangles the names of other people's playlists.
+        if owner and owner != self._backend._web_client.user_id:
+            logger.error(
+                f"Refusing to modify someone else's playlist ({playlist.uri})"
```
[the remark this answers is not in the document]
girst: spotify has ["collaborative playlists"](https://support.spotify.com/us/article/create-playlists-with-your-friends/). these can be modified by [whoever is invited to them](https://community.spotify.com/t5/Your-Library/share-and-edit-playlists-with-friends/td-p/1371906). i doubt many mopidy users are using this feature, but that was the rationale behind this wording.

that said, i've applied your suggestion.

mopidy_spotify/playlists.py
```diff
@@ -65,16 +144,114 @@ def refresh(self):
...
+                        artists = ", ".join(a.name for a in track.artists)
+                        f.write(
+                            f"#EXTINF:{length},{artists} - {track.name}\n"
+                            f"{track.uri}\n\n".encode("utf-8")
+                        )
+                logger.error(f'Created backup in "{filename}"')
+            return None
+
+        # Playlist rename logic
+        if playlist.name != saved_playlist.name:
```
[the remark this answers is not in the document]
girst: i think it makes sense to check this here, regardless of whether the api accepts it. unnamed playlists make no sense in any way.

mopidy_spotify/playlists.py
```diff
@@ -65,16 +144,114 @@ def refresh(self):
...
+        if owner and owner != self._backend._web_client.user_id:
+            logger.error(
+                f"Refusing to modify someone else's playlist ({playlist.uri})"
+            )
+            return
+
+        # We cannot add or (easily) remove spotify:local: tracks, so refuse
+        # editing if the current playlist contains such tracks.
+        if any(
+            (t.uri.startswith("spotify:local:") for t in saved_playlist.tracks)
```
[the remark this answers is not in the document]
girst: I've made a _is_spotify_local. adding it to web.linktype/web.weblink doesn't really work, since local tracks don't have ids. and modifying the class attributes to properly represent local tracks when we will never read them seems unnecessary.

mopidy_spotify/playlists.py
```diff
@@ -65,16 +144,114 @@ def refresh(self):
         self._loaded = True
 
     def create(self, name):
-        pass  # TODO
+        logger.info(f"Creating playlist {name}")
+        uri = web.create_playlist(self._backend._web_client, name)
+        self._get_flattened_playlist_refs()
```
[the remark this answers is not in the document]
girst: looks like we don't. the code i inherited had a refresh() here, so i just assumed i need to refresh the list-of-playlists. turns out, it works without in ncmpcpp and cantata, so i removed it.

mopidy_spotify/web.py
```diff
@@ -526,3 +549,91 @@ def from_uri(cls, uri):
...
+    response = web_client.post(url, json={"name": name, "public": public})
+    web_client.remove_from_cache("me/playlists")
+    return response["uri"] if response.status_ok else None
+
+
+def delete_playlist(web_client, uri):
+    playlist_id = WebLink.from_uri(uri).id
+    url = f"playlists/{playlist_id}/followers"
+    response = web_client.delete(url)
+    web_client.remove_from_cache("me/playlists")
```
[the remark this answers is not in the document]
girst: i believe we need me/playlists, so when fetching the list of playlists, the deleted one won't show up again. added the second remove, though.

mopidy_spotify/web.py
```diff
@@ -526,3 +549,91 @@ def from_uri(cls, uri):
             return cls(uri, LinkType.PLAYLIST, parts[3], parts[1])
 
         raise ValueError(f"Could not parse {uri!r} as a Spotify URI")
+
+
+def _playlist_edit(web_client, playlist, method, **kwargs):
```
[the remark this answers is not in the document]
girst: because in the test harness, `web_client_mock` is `Mock(spec=web.SpotifyOAuthClient)`. so if these functions were part of SOAC, they would be replaced by mock, and not get called, making them untestable. i didn't want to rewrite web_client_mock to use a "manually mocked" SOAC, since that would require touching a whole lot of preexisting tests (the `test_*` functions often use `.return_value` or `.side_effect`; those would have to all be replaced)

tests/test_web.py
girst: oh wow, that test was a complete brainfart, sorry for that. it is supposed to verify that non-idempotent http methods don't use the cache. rewritten and works now (also counterchecked that `.get` fails at this position.

mopidy_spotify/playlists.py
```diff
@@ -65,16 +144,114 @@ def refresh(self):
...
+            else:
+                self._patch_playlist(saved_playlist, operations)
+        except RuntimeError as e:
+            logger.error(f"Failed to save Spotify playlist: {e}")
+            # In the unlikely event that we used the replace strategy, and the
+            # first PUT went through but the following POSTs didn't, we have
+            # truncated the playlist. At this point, we still have the playlist
+            # data available, so we write it to an m3u file as a last resort
+            # effort for the user to recover from.
+            if ops_replace < ops_patch:
```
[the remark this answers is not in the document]
girst: > Why does it only get used in the replace strategy case?

because truncation can only happen when we use the replace strategy. i don't think its important to make a backup when patching the playlist. in the latter case, the user will see an error that their change failed, and the playlist is unmodified. but when using replace (and the very specific situation in the code comment happens; see last paragraph), the old playlist is clobbered, and we back it up in that case.

> And can't we do this for both playlists?

not sure what 'both'/other playlist you mean. the new state? this feature was only intended to avoid permanent data loss of previously existing data, hence i didn't implement saving the new state. (but see last big paragraph for more info)

>(why do we need length and artists? 

afaik, m3u-ext requires the length, and filling it in doesn't cost us anything. artist (and the whole extm3u block itself) were added to make the file human-parsable to an extent in this emergency situation.

>Should use Path. 

i think i am with `Extension.get_data_dir()`?

> Why binary mode? Should be ".m3u8"

i believe a remnant of a previous iteration. will change.

> and still needs a manual step to move the backup so Mopidy-M3U can find it. Could we consider something simpler? Maybe just dump (pickle?) both saved_playlist and playlist to disk? And then potentially adding a Spotify playlist restoration Command later?

i believe he likeliness of this code ever being called is extremely low. for it to trigger\*, we first have to make a successful PUT request, and then fail at a later POST request to the same endpoint, with the same authentication scope and the same parameters. and in normal use, replace mode will not be used at all, since no client supports batching multiple edit operations anyways (for a single edit, patch mode will always be used, since it's safer and requires the same amount of api requests). this is not something the user should have to deal with on a regular basis, so 'comfort features' are not included. consider this feature due diligence.

\* i'm actually creating the backup file not just when we truncate the playlist, but everytime something goes wrong with replace mode, since number of requests made is not available at this level. given that the code is pretty much never going to get called anyways, i didn't bother tightening down the check.

girst:
first off, thanks for the huge review! i've applied most of your suggested fixes (pending a black/flake pass), but i'm running out of steam today for the rest. 

regarding your initial message:

>We normally do not go for the style of mocking out the API functionality like this. I understand it results in less boilerplate/repetition in the actual tests but it's more complex and I think we should stick to the Responses-based approach we've used so far.

There are quite a lot of tests verifying the correctness of the diff calculation (and every one of them is testing some other potential bug); generating Responses for each of them seems very tedious. and given that the main point in those tests is verifying that the diffing works, won't be very useful either. that said, i'm fine with someone else writing some extra tests using Responses.

> I am also curious as to why the playlist tests run so slowly.

calculating the diff, many times over for each test.

>I will push some commits to your branch (I think that is probably the best way) over the next couple of days.

please do!

mopidy_spotify/playlists.py
```diff
@@ -56,9 +56,15 @@ def _get_playlist(self, uri, as_items=False, with_owner=False):
...
-        return xs[:i], xs[i:]
+    def _split_ended_movs(value, movs):
+        def _span(p, xs):
+            # Returns a tuple where first element is the longest prefix
+            # (possibly empty) of list xs of elements that satisfy predicate p
+            # and second element is the remainder of the list.
+            i = next((i for i, v in enumerate(xs) if not p(v)), len(xs))
+            return xs[:i], xs[i:]
+
+        return _span(lambda e: e[0] < value, movs)
```
kingosticks, line 67: I don't see the point in having to think about a lamdba in order to understand this function, but OK, at least it should be easier to test now.

mopidy_spotify/web.py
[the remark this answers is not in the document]
kingosticks: Yes - both are needed.

kingosticks:
> There are quite a lot of tests verifying the correctness of the diff calculation (and every one of them is testing some other potential bug); generating Responses for each of them seems very tedious. and given that the main point in those tests is verifying that the diffing works, won't be very useful either. 

I agree, responses at this level would be unpleasant and also not useful. But testing multiple things isn't ideal. I think we can restructure a little and test at lower levels to avoid writing too many Responses and hopefully also avoid a Python version of the Web API. It may end up being more boilerplate but I'll have a go and see where I land. Either way, doing this will probably help me review.

mopidy_spotify/utils.py
```diff
@@ -33,3 +35,71 @@ def time_logger(name, level=TRACE):
...
+        a=old, b=new, autojunk=False
+    ).get_opcodes():
+        if tag in ("insert", "replace"):
```
kingosticks, line 82: Do I understand it correctly that for a real replace operation (i.e. NOT a move) that would cost one API request, it's still converted into an add/delete pair that costs two API requests. Is there some situation where actually they can be merged into a move, or is this just a simplification? The latter is fine for now, just trying to understand. For example:
```
old = ['0', '1', '2', '3']
new = ['0', '1', 'A', 'B']
difflib.SequenceMatcher(a=old, b=new, autojunk=False).get_opcodes()
# [('equal', 0, 2, 0, 2), ('replace', 2, 4, 2, 4)]
# Becomes:
# [<insert 2 tracks [A...B] at 2>, <delete 2 tracks [2...3] at 2>]
 

```

girst: not exactly sure what you're getting at. a "replace" operation always removes and adds some tracks, so it'll always result in two api calls. merging into mov operations happens afterwards, in the code block below.

girst: splitting "replace"s into ins/del pairs is necessary to catch the most number of movs. for example, `difflib.SequenceMatcher(a="12xy34ab", b="12ab34", autojunk=False).get_opcodes()` generates a replace and a delete; this will turn into an insert, delete, delete, and then below in a mov, delete.

kingosticks: Apologies, I invented a new Web API request method that did replacing! It's been one of those days....

mopidy_spotify/playlists.py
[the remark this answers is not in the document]
kingosticks: > because truncation can only happen when we use the replace strategy.

OK, truncation is one class of issue but isn't data loss possible in either strategy? If a replacement is required and the remove operation succeeds but the corresponding add operation fails, won't that still leave you with a half modified playlist which is neither what we had or what we wanted? A backup of `playlist` in both cases would prevent that. Am I missing something?

> not sure what 'both'/other playlist you mean. the new state? this feature was only intended to avoid permanent data loss of previously existing data, hence i didn't implement saving the new state.

Your code here makes a backup of `playlist` (the requested new state), not `saved_playlist` (the old state). A request failure somewhere can leave us in a mess with our previous dat lost. So I was suggesting saving both playlists to give the best choice when it comes to recovery. 

> afaik, m3u-ext requires the length, 

No, there's no spec (yay!) and we (Mopidy-M3U) do not require it or even use it). A list of URIs is all that is required to be useful.

 > Should use Path.

Sorry, I meant [`pathlib.Path`](https://docs.python.org/3/library/pathlib.html) like the implementation in Mopidy-M3U. 

> since no client supports batching multiple edit operations anyways 

I think we've covered this before, few clients have any sort of playlist editing. We should support what is possible with the Mopidy API, not how we think people should/might/currently use it. Someone also uses it in a way you didn't think they would/should. 

> i believe he likeliness of this code ever being called is extremely low. for it to trigger*

Or we have a quota/rate limit issue half way through, or something else unlikely. But it if can happen it will happen. 

Maybe we should remove this code from the PR and address it later if anyone in the world ever sees (and reports) a problem.

mopidy_spotify/utils.py
```diff
@@ -33,3 +35,71 @@ def time_logger(name, level=TRACE):
...
+            ops.append(op("+", new[j1:j2], i1))
+        if tag in ("delete", "replace"):
+            ops.append(op("-", old[i1:i2], i1))
+
+    # then, merge pairs of insertions and deletions into a transposition:
+    # for this, we start from the rightmost element,
+    for R in reversed(range(len(ops))):
+        # and compare against all elements on its left
```
kingosticks, line 90:
```suggestion
        if ops[R].op == 'm': continue # skip any moves we've already created
        # and compare against all elements on its left
```

mopidy_spotify/playlists.py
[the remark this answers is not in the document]
girst: >truncation is one class of issue but isn't data loss possible in either strategy?

technically yes, but in my opinion only truncation is severe enough to care about. i have no problem with dropping the strategy test; will implement. 

>So I was suggesting saving both playlists to give the best choice when it comes to recovery.

will implement. 

> A list of URIs is all that is required to be useful.

useful to a computer; but i wanted it to be as useful as possible to humans as well. if we were to end up in this state, let's be as forthcoming to the user as possible.

>Sorry, I meant [`pathlib.Path`]

and [Extension.get_data_dir](https://docs.mopidy.com/en/latest/api/ext/#mopidy.ext.Extension.get_data_dir) is returning pathlib.path. we must really be talking about different things..?

>Maybe we should remove this code from the PR and address it later if anyone in the world ever sees (and reports) a problem.

strongly disagree. we should have some defence against truncation, since it's irreversible data loss. the trade-off is a few lines of imperfect code vs possible thousands of curated songs. if i were to trust mopidy-spotify with editing my years-old playlists, and it deleted most of their contents, and the maintainers just said "sorry, but thanks for the bugreport", I'd be majorly pissed. it can still be improved after committing.

mopidy_spotify/utils.py
girst: no worries, happens to the best of us ;)

mopidy_spotify/translator.py
girst: added a test for `web_to_track`, as well as testing the pass-though for `to_playlist`

kingosticks:
Sorry, I did a few bits, hit a small issue, managed to context switch along the way and lost my momentum. Would you mind rebasing on master when you get a spare moment.

girst:
On Sun, Jan 10, 2021 at 09:25:49AM -0800, Nick Steel wrote:
>Sorry, I did a few bits, hit a small issue, managed to context switch 
>along the way and lost my momentum. Would you mind rebasing on master 
>when you get a spare moment.

thanks for those "few bits!" if there's anything i can do to get you 
back up-and-running, shout now ;)

pulled your changes in (and propagated them to the new endpoints).

kingosticks:
I've got another review in progress but it's not quite done yet. Probably less work for you if it's all at once.

BlackLight:
Hey guys, what's the current status with this PR?

kingosticks:
It's not yet merged. It's waiting on me, it's my fault. I've been meaning to dig this out for weeks and I have only a couple of days before I'm unavailable for a while. Nothing motivates like an impending deadline.

girst:
I'm still interested to get this merged; it's just that since getting it 
to work for myself, the pressure is off. also, work has been a bit more 
time and energy consuming, in the last couple of months (although i 
expect that to be over relatively soon).

On Mon, May 03, 2021 at 02:57:04PM -0700, Nick Steel wrote:
>It's not yet merged. It's waiting on me, it's my fault.

don't beat yourself up over it. we're all doing this as a hobby! if you 
don't have the time, or just plain don't feel like working on it, there 
are no hard feelings from me :)

kingosticks:
We've all spent a lot of time on this and it deserves to be finished; I think in your shoes I might have been quite annoyed so I mainly want to say I _really_ appreciate the patience and understanding here. I did lose momentum and focus on this, and have also scaled back how much time I invest in this hobby but this kind response is exactly the motivation I need. Having said that, I'm about to take on an enormous priority shift (complete with a couple of weeks off work and zero free time) so while I would love to say I'll get back into this right now but I'm not sure that's remotely feasible for a few weeks.

dsynkd:
@kingosticks any chance we could get this approved?

girst:
On Tue, Mar 22, 2022 at 10:00:44AM -0700, Nick Steel wrote:
>I spent time reviewing this and reworking the tests to fit the way we 
>had done things previously - I don't want to maintain the code as it 
>is.

if you give me some direction on what you want done differently, i'd be 
happy to work on it again :)

PR #290, opened 2020-12-04
Ignore parts of web responses with a None URI.

kingosticks:
We can't do anything sensible without a URI so also ignore anything where it is ` None`. This occurs in the album and artist fields within local Spotify tracks (from a playlist).

Without this we try and translate the artist and album fields, which isn't useful and in the case of changes in https://github.com/mopidy/mopidy-spotify/pull/287/ actually causes an exception.

### DIFF
--- a/mopidy_spotify/translator.py
+++ b/mopidy_spotify/translator.py
@@ -172,7 +172,7 @@ def valid_web_data(data, object_type):
     return (
         isinstance(data, dict)
         and data.get("type") == object_type
-        and "uri" in data
+        and data.get("uri") is not None
     )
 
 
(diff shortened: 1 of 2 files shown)


kingosticks:
It would be preferable to remove all None values from the a web response but seeing as we don't go near half the fields that seems a lot of extra work.

mopidy_spotify/translator.py
```diff
@@ -172,7 +172,7 @@ def valid_web_data(data, object_type):
     return (
         isinstance(data, dict)
         and data.get("type") == object_type
-        and "uri" in data
+        and data.get("uri", None) is not None
```
jodal, line 175: The `, None` is redundant as that is the default for `.get()`.

jodal (approved):
LGTM, with the single comment.

mopidy_spotify/translator.py
kingosticks: doh!

PR #311, opened 2021-07-10
Log when refresh of playlists starts

simeg:
I have ~1k playlists on my account so a refresh takes about 8 minutes.
The logs gives no indication what is happening and I was confused so
providing an explanation to what is happening would have been useful.

PS. Thanks for a great project! I use mopidy on an RPi and connect it to a bluetooth speaker to always have music playing in my bathroom 🎶


jodal (approved):
Makes a lot of sense to log in both ends when it takes so much time. Thanks!

PR #312, opened 2021-08-01
Improve startup performance by deferring playlist loading

djmattyg007:
This change prevents loading of Spotify playlists from blocking other
common Mopidy operations. These include startup and browsing, as well as
refreshing of playlists in general. Loading Spotify playlists will
always happen in a dedicated thread, and only one refresh thread will
run at a time.

The playlists_loaded event is manually re-dispatched when Spotify is
done loading playlists, as a signal that clients should refresh their
list of playlists. This is necessary, as otherwise clients relying
solely on websocket notifications may never discover the newly-loaded
playlists.

### DIFF
--- a/mopidy_spotify/playlists.py
+++ b/mopidy_spotify/playlists.py
@@ -1,6 +1,8 @@
 import logging
+import threading
 
 from mopidy import backend
+from mopidy.core import listener
 
 import spotify
 from mopidy_spotify import translator, utils
@@ -15,6 +17,8 @@ class SpotifyPlaylistsProvider(backend.PlaylistsProvider):
         self._backend = backend
         self._timeout = self._backend._config["spotify"]["timeout"]
         self._loaded = False
+
+        self._refreshing = False
 
     def as_list(self):
         with utils.time_logger("playlists.as_list()", logging.DEBUG):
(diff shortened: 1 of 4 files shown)


kingosticks:
Sorry, I seemed to have missed this. I went down this path myself and I scrapped it. I think my main issue was reasoning about how it fitted into the (long-overdue) playlist modification PR. And it appeared quite confusing for non-webclient clients (e.g. MPD) who would see zero playlists. I was leaning towards just keeping a peristent cache of the playlist data (like libspotify does/did) and make the refreshing really quick.

djmattyg007:
Isn't it reasonable to expect non-websocket clients to have some kind of refresh functionality? Otherwise, how are they ever notified of updates? Surely if you were expecting to see playlists and saw none, you'd just refresh.

djmattyg007:
Also I'd wager that most of the time Mopidy servers are running for days or weeks without being restarted. This just impacts startup performance, a one-time cost. This is actually mostly useful for Mopidy local development.

mopidy_spotify/playlists.py
```diff
@@ -52,19 +56,36 @@ def _get_playlist(self, uri, as_items=False):
...
-                count = count + 1
-            logger.info(f"Refreshed {count} Spotify playlists")
+        logger.info("Refreshing Spotify playlists")
 
-        self._loaded = True
+        def refresher():
+            try:
+                with utils.time_logger("playlists.refresh()", logging.DEBUG):
+                    _sp_links.clear()
+                    self._backend._web_client.clear_cache()
```
3nprob, line 70: What's the motivation for explicitly clearing cache here? Seem to come from 6dd39cf14252b517bc8fe972d0adf1d4f4d6bdb3 @kingosticks

3nprob, line 70: What's the motivation for explicitly clearing cache here? Seem to come from 6dd39cf14252b517bc8fe972d0adf1d4f4d6bdb3 @kingosticks

kingosticks: Because otherwise you might still be < max-age and therefore won't fetch the fresh data, possibly leaving you with stale data. Which you don't want because you specifically asked to refresh.

3nprob: Git it - my thinking is that the ideal behavior would be to always fetch data, but keep the cache until fresh data has arrived.

Does that make sense?

mopidy_spotify/playlists.py
```diff
@@ -52,19 +56,36 @@ def _get_playlist(self, uri, as_items=False):
...
+                    count = 0
+                    for playlist_ref in self._get_flattened_playlist_refs():
```
3nprob, line 72: @djmattyg007 How much have you been running this branch? I haven't validated this properly but looks to me like there may be concurrency issues here if a client attempts to enumerate the playlists in the middle of this operation (concurrent enumeration of a collection as it's modified).

Would it make sense to introduce a mutex here?

3nprob, line 72: @djmattyg007 How much have you been running this branch? I haven't validated this properly but looks to me like there may be concurrency issues here if a client attempts to enumerate the playlists in the middle of this operation (concurrent enumeration of a collection as it's modified).

Would it make sense to introduce a mutex here?

kingosticks:
Thanks for the great original work on this. Sorry it took so long to get the improvement in. Closing in favour of https://github.com/mopidy/mopidy-spotify/pull/382 which is based, in part, on this.

Issue #297, opened 2021-01-04
Liked tracks


winterscar:
The main way I use Spotify is to navigate to my "Liked Tracks" and hit "Play All".
I use this library in association with home-assistant, and I can start playing any of my playlists without problems, but I'm unable to figure out how I might instruct mopidy to just shuffle all my "Liked Tracks" from spotify.

Is this possible using this library?

Thanks.

jodal:
For all music backends, Mopidy supports both:
- physically shuffling the play queue (called the "tracklist") so that you can view what order the tracks will be played in,
- "random" play mode, where it will pick the next song by random.

I'm not familiar with the HomeAssistant Mopidy integration, so I don't know if or how this functionality is exposed there. Please try asking at https://discourse.mopidy.com or the maintainers of the HomeAssistant integration.

winterscar:
Hi, thanks for clarifying about the shuffle modes mopidy supports. However, my question was more around how I can get my "liked songs" from spotify into the tracklist in the first place.

jodal:
AFAIK we don't expose "Liked Songs" in any way, and I can't immediately find any support for them in [Spotify's Web API](https://developer.spotify.com/documentation/web-api/reference/). If anyone can point out my error or if Spotify adds support for it to the Spotify Web API, we can make "Liked Songs" available in Mopidy-Spotify too.

winterscar:
Would it not be this?
https://developer.spotify.com/documentation/web-api/reference-beta/#endpoint-get-users-saved-tracks

djmattyg007:
I think @winterscar is referring to the "folders" presented through the `browse` integration, by navigating to Spotify -> Your music -> Your Tracks. This isn't a playlist and therefore presumably can't be triggered through an integration that only supports playlists.

It might be possible to provide a fake "liked tracks" playlist?

winterscar:
That seems like a good solution to me!

djmattyg007:
@winterscar Would you be able to take a look at #313 and see if it resolves your issue?

PR #313, opened 2021-08-09
Support lookups for 'your' URIs

djmattyg007:
This adds the ability to utilise the "spotify:your:tracks" and
"spotify:your:albums" URIs through library lookups.

### DIFF
--- a/mopidy_spotify/lookup.py
+++ b/mopidy_spotify/lookup.py
@@ -1,7 +1,8 @@
 import logging
 
 import spotify
-from mopidy_spotify import playlists, translator, utils, web
+from mopidy_spotify import browse, playlists, translator, utils
+from mopidy_spotify.web import LinkType, WebLink
 
 logger = logging.getLogger(__name__)
 
@@ -12,16 +13,18 @@ _VARIOUS_ARTISTS_URIS = [
 
 def lookup(config, session, web_client, uri):
     try:
-        web_link = web.WebLink.from_uri(uri)
-        if web_link.type != web.LinkType.PLAYLIST:
+        web_link = WebLink.from_uri(uri)
+        if web_link.type not in (LinkType.PLAYLIST, LinkType.YOUR):
             sp_link = session.get_link(uri)
     except ValueError as exc:
         logger.info(f"Failed to lookup {uri!r}: {exc}")
         return []
 
     try:
-        if web_link.type == web.LinkType.PLAYLIST:
+        if web_link.type == LinkType.PLAYLIST:
             return _lookup_playlist(config, session, web_client, uri)
+        elif web_link.type == LinkType.YOUR:
+            return list(_lookup_your(config, session, web_client, uri))
         elif sp_link.type is spotify.LinkType.TRACK:
             return list(_lookup_track(config, sp_link))
         elif sp_link.type is spotify.LinkType.ALBUM:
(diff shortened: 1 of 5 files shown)


djmattyg007:
I believe this should resolve #297

marnunez:
It does! Thanks!

kosciej16:
@djmattyg007 Is there a chance to merge it? Possibility of playing liked songs would be awesome!

mopidy_spotify/lookup.py
```diff
@@ -92,3 +95,30 @@ def _lookup_playlist(config, session, web_client, uri):
...
+                web_track, bitrate=config["bitrate"]
+            )
+            if track is not None:
+                yield track
+    elif variant == "albums":
+        for item in items:
+            # The extra level here is to also support "saved album objects".
+            web_album = item.get("album", item)
+            if not translator.valid_web_data(web_album, "album"):
+                continue
```
kingosticks, line 121:
```suggestion
            album_ref = translator.web_to_album_ref(web_album)
            if album_ref is None:
                continue
```

mopidy_spotify/lookup.py
```diff
@@ -92,3 +95,30 @@ def _lookup_playlist(config, session, web_client, uri):
...
+
+
+def _lookup_your(config, session, web_client, uri):
+    parts = uri.replace("spotify:your:", "").split(":")
+    if len(parts) != 1:
+        return
+    variant = parts[0]
+
+    items = browse._load_your_music(web_client, variant)
+    if variant == "tracks":
```
kingosticks, line 107: One thing different between the two varients here is that tracks won't be loaded by libspotify whereas albums will. I wonder if the difference is observable

kingosticks (approved):

mopidy_spotify/lookup.py
```diff
@@ -92,3 +95,30 @@ def _lookup_playlist(config, session, web_client, uri):
...
+            )
+            if track is not None:
+                yield track
+    elif variant == "albums":
+        for item in items:
+            # The extra level here is to also support "saved album objects".
+            web_album = item.get("album", item)
+            if not translator.valid_web_data(web_album, "album"):
+                continue
+            sp_link = session.get_link(web_album.get("uri"))
```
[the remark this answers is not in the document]
djmattyg007: Would you be able to explain what the meaningful difference is between the two versions of this code?

mopidy_spotify/lookup.py
djmattyg007: Would you be able to elaborate here? I'm pretty sure I tested this quite extensively at the time I wrote it. Unless something has changed with some other component, it should work.

kingosticks: When you load any sort of container thing in libspotify it loads the track data for that in the background so if you then go and use the track (I.e. play it) it should happen instantly. If the tracks are being loaded via the Web API, when you go to play it that'll be the first time libspotify hears about the track and first has to load the data. There might be a short pause for that but I wasn't sure it'd be noticeable.

mopidy_spotify/lookup.py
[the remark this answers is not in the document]
kingosticks: Using the existing function hides the internal Web API structure and does some error checking for you. I appreciate you still need to know how the response is structured later on but this is one less place. I also have so far preferred not to called `valid_web_data` outside of the translator (although maybe we do this already, I can't remember and I've not checked)

mopidy_spotify/lookup.py
djmattyg007: Is there an alternative that would avoid that potential pause, and would it be worth scrapping what I've worked on in this PR to implement it vs merging this now and implementing that later?

kingosticks: The only way I know (and it's a little hacky) is to do something like [this](https://github.com/mopidy/mopidy-spotify/blob/master/mopidy_spotify/playlists.py#L98-L110). I'll test it myself later but if you didn't notice any issue it's probably fine. I'd be happy to merge as it is and we can always come back later if people find the performance is an actual problem.

djmattyg007: I did wonder what the purpose of that code was.

Either way, I definitely think this feature is better off existing in a non-perfect state than not existing at all. I'll make the translator-related change you suggested to the lookup code at some point over the next couple of days, and then we can get this merged :+1:

djmattyg007:
I've updated this PR. It should be good to go :+1:

kingosticks:
Cool, so as a resulf of this we can now do:
* `core.library.lookup(uris=["spotify:your:tracks"])`
* `core.library.search(query={"uris": ["spotify:your:tracks"]})`

e.g.
```
curl -d '{"jsonrpc": "2.0", "id": 1, "method": "core.library.lookup", "params": {"uris": ["spotify:your:tracks"]}}' -H 'Content-Type: application/json' http://localhost:6680/mopidy/rpc

curl -d '{"jsonrpc": "2.0", "id": 1, "method": "core.library.search", "params": {"query": {"uri": ["spotify:your:tracks"]}}}' -H 'Content-Type: application/json' http://localhost:6680/mopidy/rpc
```

Perhaps we can use this functionality when we fix https://github.com/mopidy/mopidy-spotify/issues/328 (in a later PR).

Issue #259, opened 2020-03-30
Login is only attempted when mopidy starts


Gigahawk:
If mopidy is started without an internet connection, the Spotify login will fail, and no attempt to reconnect is made even if the connection is restored.

The plugin should periodically attempt to login if it is currently not logged in.

kingosticks:
Alternatively, could you ensure you wait for your Internet connection to become ready before starting Mopidy? The Mopidy systemd unit waits for the network target.

Gigahawk:
I'm on a distro that doesn't use systemd, I just have mopidy start with my window manager.

s-martin:
We use mopidy-spotify in our project and users have reported just recently that login to Spotify doesn't work anymore. Reason could be, that internet connection is not available yet, but this was never an issue in the past.
See:
https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/893
https://github.com/MiczFlor/RPi-Jukebox-RFID/issues/901

Were there any changes in 3.0.2?

kingosticks:
The changelog is available online but there were no changes in Mopidy 3.0.2 that would impact the ability of libspotify to login.

s-martin:
> The changelog is available online but there were no changes in Mopidy 3.0.2 that would impact the ability of libspotify to login.

Thanks for the info and pointing us into the right direction.

kobbejager:
I have the same issue on Rasbian Buster Lite. In /etc/init.d/mopidy network is listed as a requirement, but apparently it does not wait for the (wired) network to be fully started. Anyway, I agree with the initial comment that Mopidy-Spotify should try to reconnect regularly.

jaor:
fwiw, same for me with debian sid and mopidy started using the systemd
unit on boot.  i'm never logged in spotify when i first access mopidy
after a reboot (the server is running, and other backends working).  i
need to restart via service for logging to happen.

kingosticks:
Mopidy requires the network and we already specify that in our service files. If your system isn't waiting for the network to be ready, I'm finding it hard to see how that's a Mopidy issue. I've already mentioned the fix for Raspbian above. 

I get there may be some other case that I don't appreciate and so if someone wants to submit a PR for this (that works properly and is tested) then I will review it.

PR #344, opened 2022-06-16
Make thread-safe, retain cache despite refresh

### LINKED ISSUE #296: fail to login at startup
Hi,
I'm using mopidy_spotify v4.0.1 on armbian buster on nano pi NEO SBC. I see there is an issue with spotify login when login fails on mopidy startup. For example, when there is no internet connection at mopidy startup and the internet connection comes back after couple of minutes spotify backend does not try to login again.
To reproduce the issue, the connection must not come back until this log line appears:
```
ERROR [24188:SpotifyEventLoop] spotify.session: Spotify login error: <ErrorType.UNABLE_TO_CONTACT_SERVER: 8>
```
I have a solution but i'm not sure if this is correct, but at least it solves the problem for me:
```
diff --git a/mopidy_spotify/backend.py b/mopidy_spotify/backend.py
index 7cefb48..e4777dc 100644
--- a/mopidy_spotify/backend.py
+++ b/mopidy_spotify/backend.py
@@ -90,6 +90,16 @@ class SpotifyBackend(pykka.ThreadingActor, backend.Backend):
             self._logged_out,
             backend_actor_proxy,
         )
+        def logged_in(session, error_type):
+            if error_type is spotify.ErrorType.OK:
+                logger.error('Logged in as %s' % session.user)
+            else:
+                logger.error('Login failed: %s' % error_type)
+                session.login(
+                    self._config["spotify"]["username"],
+                    self._config["spotify"]["password"],
+                )
+        session.on(spotify.SessionEvent.LOGGED_IN, logged_in)
         session.on(
             spotify.SessionEvent.PLAY_TOKEN_LOST,
             on_play_token_lost,
```

3nprob:
Attempt at making #312 thread-safe.

Also:
* Add `force_refresh` parameter to webclient and use it to remove the need for cache-wiping on refresh
* Retry the entire refresh process indefinitely  (this is separate from the per-request exponential backoff), with a 3s interval. (resolves #296, #259 )
* Minor code cleanup

### DIFF
--- a/mopidy_spotify/playlists.py
+++ b/mopidy_spotify/playlists.py
@@ -27,11 +27,11 @@ class SpotifyPlaylistsProvider(backend.PlaylistsProvider):
 
             return list(self._get_flattened_playlist_refs())
 
-    def _get_flattened_playlist_refs(self):
+    def _get_flattened_playlist_refs(self, force_refresh=True):
         if not self._backend._web_client.logged_in:
             return []
 
-        user_playlists = self._backend._web_client.get_user_playlists()
+        user_playlists = self._backend._web_client.get_user_playlists(force_refresh=force_refresh)
         return translator.to_playlist_refs(
             user_playlists, self._backend._web_client.user_id
         )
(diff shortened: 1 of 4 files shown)


kingosticks (commented):
Let's pause on this until pyspotify is out the way.

kingosticks:
Sorry, I totally forgot to take a look. Will do this week.

kingosticks:
It took me a while... for which I am sorry but this was useful and I based https://github.com/mopidy/mopidy-spotify/pull/382 on this. Closing in favour of the updated PR.

PR #345, opened 2022-06-16
tests: fix responses match_querystring deprecation

3nprob:
* Fix test not checking querystring anymore (https://github.com/getsentry/responses/pull/432)
* Fix error logging fail if result is None


kingosticks:
These fixes went in via other PRs, not sure why we left this open.

PR #354, opened 2023-04-03
Use python librespot port

beaverking1212:
Use librespot-python in place of gst-plugins-spotify

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -89,35 +89,7 @@ Dependencies
   Settings > Security > App passwords > Generate app passwords, and generate one
   to use with Mopidy-Spotify.
 
-- ``gst-plugins-spotify`` >= 0.10. The `GStreamer Rust Plugin
-  <https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs>`_ to stream Spotify
-  audio, based on `librespot <https://github.com/librespot-org/librespot/>`_.
-  This plugin is not yet available from apt.mopidy.com and must be built from
-  source (see below).
-
 - ``Mopidy`` >= 3.4. The music server that Mopidy-Spotify extends.
-
-Example build instructions for ``gst-plugins-spotify``::
-
-1. `Install Rust <https://www.rust-lang.org/tools/install>`_::
-
-    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
-
-2. Install the `GStreamer Rust bindings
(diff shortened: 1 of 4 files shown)


kingosticks:
I'm not sure about this. At a high level, before I even look at this code, using gst-plugin-spotify has benefits:

* Passing over all the heavy lifting to a GST element that's maintained by the core GST team. 
* Rust *should* perform better. The performance of Mopidy-Spotify on a pi0 has never been great and a Python version can only be worse.

From a non-technical aspect, one of librespot-python's examples is a streaming server; that's an explicit Spotify TOS "no" and great way to get their project banned. They've made it far too easy for their software to be abused. I don't want to depend on something like that. I also personally think that "librespot" is a daft PyPI project name for them to use.

Once we get a gst-plugin-spotify packaged, I don't think there is any benefit to using librespot-python. Thoughts?

beaverking1212:
This solution uses only Python code with no non-Python dependencies and no system-level software installed.

Unless we have a soon ETA for packaging gst-plugin-spotify and it will support all major architectures, I think this is much more realistic solution.

Maybe it should be configurable whether gst-plugin-spotify or librespot-python is used?

beaverking1212:
I think making unrealistic plans, having no capacity to implement them and _blocking the alternatives_ is a huge problem for several open-source projects, including some Mopidy related.

kingosticks:
I appreciate the effort but this is the opposite direction to what we've all been discussing so far. I think there's far more value in depending on a robust GStreamer solution that someone else maintains. Currently that means we need to package + provide that dependency and that effort has been slow. In the future, the GST Rust plugins will be officially packaged for Debian etc and hopefully we can encourage them to include the new spotify plugin. 

Mopidy requires a GStreamer install and Mopidy-Spotify has historically always had an extra non-python dependency, which I think we've shown isn't a problem. Supporting two different versions is even more work (for our imaginary maintainer) so I think that's unwise. We've previously run into problems building on/with GStreamer's file-like elements, they don't play well with others and in our experience are best avoided.

A brief note/discussion first would have been preferable to dropping this here unannounced and without context. Even if this PR was ready to merge (it isn't), it's not a direction I'm interested in spending my time on. Instead, myself and other contributors  have been testing the gst-plugin-spotify solution and although it's still missing seeking, I think it's otherwise good. There's plenty of room for someone to come along and sort out packaging across the archs we all want. There has been some helpful stuff but nothing resembling a PR yet. That would be a useful thing to help with.

Issue #353, opened 2023-03-27
"New login to Spotify" emails


kingosticks:
I am *sometimes* getting emails from Spotify notifying my of new logins when playing songs with the latest version. My understanding was these would not be sent when using cached credentials so something is going wrong. The credentials file does exist and looks sensible. The current logging isn't really helping to understand the problem.

> New login to Spotify
> 
> We noticed you logged in from a new device. If this was you, there’s nothing for you to do right now.
>
> Location
> Time March 27, 2023 at 4:05:38 PM BST

beaverking1212:
This needs fixes in gst-plugin-spotify and credentials.json must not be deleted by file cache cleanup.

I see two options to protect credentials.json
Option 1 is storing it in a folder under home directory.
Option 2 is storing file cache in in subfolder of the cache directory. This is similar to what libspotify did, but this will prevent old cache files from being cleaned up by cache cleanup.

@kingosticks What do you think?

kingosticks:
Are you saying that the problem is the spotifysrc deleting its own credentials.json file accidentally? When is it doing the cleanup? My theory was different.

Either way, I don't think option 1 works when running as a system service. 

Maybe I don't quite understand the problem but is there an option 3 is to put credentials.json in a subfolder?

beaverking1212:
> Are you saying that the problem is the spotifysrc deleting its own credentials.json file accidentally? When is it doing the cleanup?

No. Currently problem is in gst-plugin-spotify, but when that will be fixed cache cleanup may become next problem.

I have tested option 1 and option 2 on Raspberry Pi OS and they both work.
Subfolders of file cache are also cleaned up by the cache cleanup, so option 3 will not help.

beaverking1212:
May be option 4 would be creating directory under [Extension.get_data_dir()](https://docs.mopidy.com/en/latest/api/ext/#mopidy.ext.Extension.get_data_dir)? May be this is the best?

Besides should credentials be cached if allow_cache is off? I think yes because it is part of login and not file cache.

kingosticks:
Correct me if wrong but are we looking at two different issues here.

1. The one I reported is the annoying new login emails. I think this can be avoided with a change to gst-plugin-spotify, having it cache the "reusable credentials" like normal rather than the manual steps it does to save the original credentials, for some reason.

2. The credentials.json cache file being erroneously removed. I've never had this, but sounds like you have?

There's another related one:

[...]

PR #357, opened 2023-04-21
Improve credentials cache

beaverking1212:
Always cache credentials and use dedicated folder for credentials cache. Folder for credentials cache has mode 700.
For #353

### DIFF
--- a/mopidy_spotify/backend.py
+++ b/mopidy_spotify/backend.py
@@ -1,5 +1,3 @@
-import os
-
 import pykka
 from mopidy import backend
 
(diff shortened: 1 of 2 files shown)


mopidy_spotify/backend.py
```diff
@@ -37,13 +39,18 @@ class SpotifyPlaybackProvider(backend.PlaybackProvider):
...
         self._config = self.backend._config["spotify"]
 
     def on_source_setup(self, source):
         for prop in ["username", "password", "bitrate"]:
             source.set_property(prop, str(self._config[prop]))
+        credentials_dir = os.path.join(self._data_location, "credentials-cache")
+        if not os.path.exists(credentials_dir):
+            os.makedirs(self._data_location, exist_ok=True)
+            os.mkdir(credentials_dir, 0o700)
+        source.set_property("cache-credentials", credentials_dir)
```
kingosticks, line 52: We try and use Pathlib in new code e.g.

```suggestion
        credentials_dir = self._data_location / "credentials-cache"
        if not credentials_dir.exists():
            credentials_dir.mkdir(mode=0o700, exist_ok=True)
        source.set_property("cache-credentials", credentials_dir)
```

kingosticks: We could also move this to `SpotifyPlaybackProvider.__init__()` (or even `SpotifyBackend.on_start()`) since it's really just a one-off concern.

kingosticks (changes requested):

beaverking1212:
Fixed

`Extension().get_data_dir(self.backend._config)` makes the folder it returns. Now folders are created on Mopidy start.

kingosticks:
Brilliant, thanks.

PR #382, opened 2024-03-12
Fix startup time

kingosticks:
I think this is the best of both worlds since it gets the playlist names immediately, avoids blocking Mopidy startup, but still gathers the track info in the background to make everything snappy later on.

### DIFF
--- a/src/mopidy_spotify/playlists.py
+++ b/src/mopidy_spotify/playlists.py
@@ -13,22 +13,17 @@ class SpotifyPlaylistsProvider(backend.PlaylistsProvider):
     def __init__(self, backend):
         self._backend = backend
         self._timeout = self._backend._config["spotify"]["timeout"]
-        self._loaded = False
-
-        self._refreshing = False
+        self._refresh_mutex = threading.Lock()
 
     def as_list(self):
         with utils.time_logger("playlists.as_list()", logging.DEBUG):
-            if not self._loaded:
-                return []
-
             return list(self._get_flattened_playlist_refs())
 
-    def _get_flattened_playlist_refs(self):
+    def _get_flattened_playlist_refs(self, *, refresh=False):
         if not self._backend._web_client.logged_in:
             return []
 
-        user_playlists = self._backend._web_client.get_user_playlists()
+        user_playlists = self._backend._web_client.get_user_playlists(refresh=refresh)
         return translator.to_playlist_refs(
             user_playlists, self._backend._web_client.user_id
         )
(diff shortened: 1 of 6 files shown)


kingosticks:
Do with a bit more testing with some other clients to ensure behaviour makes sense.

src/mopidy_spotify/web.py
```diff
@@ -451,7 +463,9 @@ def _with_all_tracks(self, obj, params=None):
         track_pages = self.get_all(
             tracks_path,
             params=params,
-            ignore_expiry=obj.status_unchanged,
+            expiry_strategy=ExpiryStrategy.FORCE_FRESH
+            if obj.status_unchanged
+            else None,
```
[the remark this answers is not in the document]
kingosticks: Thanks, I really didn't like what it did here

src/mopidy_spotify/playlists.py
```diff
@@ -13,6 +15,8 @@ def __init__(self, backend):
         self._timeout = self._backend._config["spotify"]["timeout"]
         self._loaded = False
 
+        self._refreshing = False
```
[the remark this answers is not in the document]
adamcik: I would use the lock as jodal suggested, otherwise you'll need to have a something protecting `_refreshing` from concurrent access. And if you don't go that route I would annotate with comments saying something about what lock needs to be held for it to be safe to read/write this variable.

(This might be all my C++ at work bleeding over, but the concurrency issues should still be the same)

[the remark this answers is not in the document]
adamcik:
```
    def __init__(self, backend):
        self._refresh_mutex = threading.Lock()

    def refresh(self):
        if not self._refresh_mutex.acquire(blocking=False):
            logger.info("Refreshing Spotify playlists already in progress")
            return
        ...

    def _refresh_tracks(self, playlist_uris):
        if not self._refresh_mutex.locked():
            logger.error("Lock must be held before calling this method")
            return

        try:
            ...
        except Exception:
            logger.exception("Error occurred while refreshing Spotify playlists tracks")
        finally:
            self._refresh_mutex.release()
```

src/mopidy_spotify/playlists.py
```diff
@@ -48,18 +47,32 @@ def _get_playlist(self, uri, *, as_items=False):
...
+            logger.exception("Error occurred while refreshing Spotify playlists")
+
+    def _refresh_tracks(self, playlist_uris):
+        self._refreshing = True
+        try:
+            with utils.time_logger("playlists._refresh_tracks()", logging.DEBUG):
+                refreshed = [uri for uri in playlist_uris if self.get_items(uri)]
+                logger.info(f"Refreshed {len(refreshed)} Spotify playlists")
+
+            listener.CoreListener.send("playlists_loaded")
```
adamcik, line 71: Is it safe to call this from a different thread/actor?

src/mopidy_spotify/playlists.py
```diff
@@ -48,18 +47,32 @@ def _get_playlist(self, uri, *, as_items=False):
...
+                daemon=True,
+            ).start()
+        except Exception:
+            logger.exception("Error occurred while refreshing Spotify playlists")
+
+    def _refresh_tracks(self, playlist_uris):
+        self._refreshing = True
+        try:
+            with utils.time_logger("playlists._refresh_tracks()", logging.DEBUG):
+                refreshed = [uri for uri in playlist_uris if self.get_items(uri)]
```
adamcik, line 68: We could also just use the length of the uris from the input?

src/mopidy_spotify/web.py
```diff
@@ -9,6 +9,7 @@
 from email.utils import parsedate_to_datetime
 from enum import Enum, unique
 from http import HTTPStatus
+from threading import Lock
```
adamcik, line 12: Personally I would prefer just importing threading here, but if this better matches the current code style than leave it as is. Reasoning being that having the module in there makes it clearer which/what Lock this is.

src/mopidy_spotify/web.py
```diff
@@ -64,6 +65,8 @@ def __init__(  # noqa: PLR0913
 
         self._headers = {"Content-Type": "application/json"}
         self._session = utils.get_requests_session(proxy_config or {})
+        self._cache_mutex = Lock()
```
adamcik, line 68: What do these protect? It probably doesn't hurt to have a comment by each variable/method that require the mutex to be held saying `Can only be accessed/called while _cache_mutex is held` etc

src/mopidy_spotify/playlists.py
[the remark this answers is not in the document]
kingosticks: I was being lazy and relying on the fact this is atomic in (current versions of) Python. I should just use a lock and make this interaction explicit.

src/mopidy_spotify/playlists.py
```diff
@@ -48,18 +47,32 @@ def _get_playlist(self, uri, *, as_items=False):
...
-
-        self._loaded = True
+        if self._refreshing:
+            logger.info("Refreshing Spotify playlists already in progress")
+            return
+        try:
+            uris = [ref.uri for ref in self._get_flattened_playlist_refs(refresh=True)]
+            logger.info(f"Refreshing {len(uris)} Spotify playlists in background")
+            threading.Thread(
+                target=self._refresh_tracks,
```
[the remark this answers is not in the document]
kingosticks: Originally I thought the same and I started down that road. It was a lot of code changes to resolve the futures everywhere (albeit simple) but my main gripe was it seemed a waste to add overhead (moving between the threads) everywhere for handling just this one async case. So I tore it all out and went with the threaded version from the existing PR.

src/mopidy_spotify/playlists.py
kingosticks: Yes, it `tell`s other pykka actors about the event.

src/mopidy_spotify/playlists.py
kingosticks: I was trying to filter out playlists that didn't refresh successfully with the ` if self.get_items(uri)`. Originally I used `_get_playlist` here instead but for some reason changed it to `get_items`, in retrospect that was a bad change since what I have now will incorrectly also filter out playlists with no tracks.

src/mopidy_spotify/web.py
kingosticks: Yes... I think this comment is fair and exposes why this is a bit hacky... Would it helped if I claimed I was going to redo all the caching layers anyway?

adamcik: Yeah, the fact that this base class has a cache mutex which protects something in the child class is a bit off. That is based on my reading of this being that `SpotifyOAuthClient._cache` is what `OAuthClient._cache_mutex` protects, so this does feel like it needs some refactoring or cleanup of sorts.

adamcik: While the refresh case I assume is more a "single request in flight" type scenario which effectively protects `_expires` and `_headers`?

adamcik: Anyway, feel free to postpone addressing this. But perhaps add a todo or something so we reduce the risk of forgetting this?

kingosticks:
Chasing the last bit of coverage found a bug. The system works!

PR #397, opened 2024-09-22
Playback using access token

kingosticks:


adamcik (approved):
Would it make sense to leave username/password in the docs but marked as obsolete? Would that be less confusing than just removing it?

jodal (approved):

PR #434, opened 2026-06-22
fix: version fall back when package metadata is unavailable

adamcik:
I'm running things with nix these days, which runs without git, which means `version` might fail. In that case it is better to fallback to a dummy version IMO.

### DIFF
--- a/src/mopidy_spotify/__init__.py
+++ b/src/mopidy_spotify/__init__.py
@@ -1,11 +1,14 @@
 import pathlib
-from importlib.metadata import version
+from importlib.metadata import PackageNotFoundError, version
 from typing import override
 
 import cyclopts
 from mopidy import config, ext
 
-__version__ = version("mopidy-spotify")
+try:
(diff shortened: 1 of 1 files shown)


kingosticks (approved):

jodal:
I think we could consider removing `__version__` from all of our packages. `importlib.metadata.version()` is the right way to look up a package version in 2026.

adamcik:
Sure, but the problem I had is that if that is pulling from git, and we have code that calls version I get in trouble. As typically I'll setup `nix flake check` which then runs `pytest` or whatever else in a hermetic env with just minimal deps and no copy of `.git`.

But if `version()` was only being called here and nowhere else it effectively solves the problem for me

jodal:
I made https://github.com/mopidy/mopidy/pull/2298, which I intent do merge as part of 4.1 later. Once that is released, extensions requiring at least 4.1 can start dropping `__version__` and `Extension.version`, removing the need for doing workarounds like this PR in all extensions.

PR #436, opened 2026-06-23
feat: Start preparing for PKCE support

adamcik:
- Add auth-related state handling and wiring
- New path is isolated so both flows can coexist during rollout
- `auth.json` is the source of truth for the auth state machine
- It distinguishes bridge and PKCE modes
- It also captures failure states and PKCE refresh token handling
- Inlcudes a `utils.replace(...)` for atomic file replacement (with modes)


src/mopidy_spotify/utils.py
```diff
@@ -14,6 +17,7 @@
 
 if TYPE_CHECKING:
     from collections.abc import Generator, Iterable
+    from pathlib import Path
 
     from mopidy.config import ProxyConfig
```
jodal, line 21: We don't need this both at runtime and in the type checking block. Get rid of the `as PathlibPath` part here and remove the `Path` import from the type checking block.

src/mopidy_spotify/utils.py
```diff
@@ -45,6 +49,42 @@ def time_logger(name: str, level: int = TRACE) -> Generator[None]:
...
+) -> Generator[tempfile._TemporaryFileWrapper[bytes]]:
+    temp_path: PathlibPath | None = None
+    try:
+        with tempfile.NamedTemporaryFile(
+            mode="wb",
+            dir=path.parent,
+            prefix=f".{path.name}.",
+            delete=False,
+        ) as file_handle:
+            temp_path = PathlibPath(file_handle.name)
```
jodal:
```suggestion
            temp_path = Path(file_handle.name)
```

jodal (approved):
LGTM. Hoping you'll continue this work!

PR #442, opened 2026-09-18
Add Spotify PKCE authentication command

adamcik:
Add the local authorization half of Spotify PKCE without changing runtime token refresh behavior.

- Generate a PKCE verifier, challenge, and CSRF state.
- Open Spotify authorization through the Mopidy website callback.
- Accept either the full callback URL or the existing encoded result.
- Validate returned state before exchanging the authorization code.
- Apply the configured network timeout with a one-second minimum.
- Store versioned authorization state atomically in `auth.json` with mode `0600`.
- Redact refresh tokens from representations and validation diagnostics.
- Clear local authorization state during logout.
- Cover the complete command-to-file flow, including stored payload and permissions.

This is intentionally preparatory. Mopidy-Spotify continues using configured bridge credentials at runtime; this command only creates local authorization state for the follow-up provider integration.

Later stacked PRs extract refresh providers, add reusable secret storage, activate PKCE preference with bridge fallback, and make bridge credentials optional.


adamcik:
https://github.com/adamcik/mopidy-spotify/pull/3/changes has the full stack if you want to peek at the rest of it.

src/mopidy_spotify/auth_flow.py
```diff
@@ -0,0 +1,203 @@
...
+        msg = str(exc) or exc.__class__.__name__
+        raise ValueError(msg) from exc
+    except ValueError as exc:
+        msg = "invalid token response."
+        raise ValueError(msg) from exc
+
+    try:
+        result = TokenExchangeResponse.model_validate(payload)
+    except ValidationError as exc:
+        msg = "missing refresh_token."
```
jodal: Do we know that this is the correct diagnosis in this branch?

src/mopidy_spotify/auth_flow.py
```diff
@@ -192,12 +199,13 @@ def finish_auth(self, challenge: AuthChallenge, pasted_result: str) -> AuthSucce
...
-        if refresh_token is None:
+        secret = result.refresh_token
+        if secret is None:
             msg = "missing refresh_token."
             raise AuthExchangeError(msg)
+        refresh_token = secret.get_secret_value()
 
         self._auth_state_store.save(
-            auth_state.PkceAuthorizedAuthPayload(refresh_token=refresh_token)
+            auth_state.PkceAuthorizedAuthPayload(refresh_token=SecretStr(refresh_token))
```
jodal: Could this just be `=secret` instead of wrapping it again?

src/mopidy_spotify/commands.py
```diff
@@ -1,23 +1,69 @@
...
+    write_output: Callable[..., None] = print,
+) -> int:
+    challenge = flow.start_auth()
+
+    write_output(
+        "Please visit the following URL. After authorizing, paste the result "
+        "shown in your browser here:\n"
+    )
+    write_output(challenge.authorization_url)
+    write_output()
```
jodal, line 27: A bit strange to ask you to paste something after the colon, and then print an URL there.

What about something like this instead?

```py
    write_output("Please visit the following URL:\n")
    write_output(challenge.authorization_url)
    write_output()
    write_output(
        "After authorizing, paste the result "
        "shown in your browser here:\n"
    )
    write_output()
```

src/mopidy_spotify/commands.py
```diff
@@ -1,23 +1,69 @@
...
+        write_output("Authentication aborted.")
+        return 1
+
+    try:
+        flow.finish_auth(challenge, pasted_result)
+    except auth_flow.AuthFlowError as exc:
+        write_output(str(exc))
+        return 1
+
+    write_output("Refresh token successfully stored.")
```
jodal, line 44: Add an empty line before this one, so we get some distance from the pasted result.

src/mopidy_spotify/commands.py
```diff
@@ -1,23 +1,69 @@
...
+    config = Config.get_global()
+    auth_state_path = Extension.get_auth_state_path(config)
+    flow = auth_flow.AuthFlow(config, auth_state_path)
+    return run_auth_command(flow)
+
+
 @app.command(help="Logout from Spotify account.")
 def logout() -> None:
     config = Config.get_global()
     credentials_dir = Extension().get_credentials_dir(config)
```
jodal: `get_credentials_dir` is also a staticmethod, so we can drop the instantiation of the class for consistency with the new line added.

src/mopidy_spotify/commands.py
```diff
@@ -29,6 +75,10 @@ def logout() -> None:
                 dir_path.rmdir()
                 logger.debug(f"Removed directory {dir_path}")
         credentials_dir.rmdir()
+        auth_state.FileAuthStateStore(auth_state_path).save(
+            auth_state.ClearedAuthPayload(mode=mode)
+        )
+        logger.debug(f"Cleared file {auth_state_path}")
```
jodal: Any reason to wrap the credentials dir cleanup in OAuth cleanup, or could we puth the OAuth part before or after?

jodal (approved):
Tried out the new command, and it works great 🙌

src/mopidy_spotify/auth_flow.py
adamcik: True, updated to make it more correct.

src/mopidy_spotify/commands.py
adamcik: Should be better now.

src/mopidy_spotify/commands.py
adamcik: Done

src/mopidy_spotify/commands.py
adamcik: Done

src/mopidy_spotify/commands.py
adamcik: fixed

src/mopidy_spotify/auth_flow.py
adamcik: fixed

kingosticks:
Somewhat related, gstspotifysrc merged an auth helper binary recently. We need a couple of changes to add it to our deb package.

The helper uses OAuth (using Spotify's desktop client ID) to easily obtain the credentials required for librespot playback. We could add another Mopidy-Spotify command to call that helper, or even integrate it into this one?

adamcik:
Could be, I'm just blindly focusing on getting PKCE in, pretending the other stuff isn't even happening. I should probably take a closer look at the other things and see how/if we can unify it

adamcik:
I looked through the PKCE stack. I suggest shaping the commands as:

```
mopidy spotify auth web
mopidy spotify auth playback
mopidy spotify auth status
```

Bare auth can show help for now and later run both flows. Web PKCE and playback credentials should remain separate because neither token works for both APIs. Initially, playback can invoke gstspotify-auth with Mopidy’s credentials directory; tighter library integration can come later. We should also stop telling users to delete credentials.json and consider no longer passing the unusable Web API token to spotifyaudiosrc?

